PR #78, opened 2013-03-20
Add an origin column to the database that is set in configuration

jrivord:
The reason for this change is to enable you to have a tag in the configuration file called “Origin” that will then be placed in the row of the database. The system supports the class and line, but no way to tell what application generated a message. If you have three applications all writing to the same database log there is no good way to distinguish which application generated which message.

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/db/DBAppender.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/db/DBAppender.java
@@ -45,6 +45,7 @@ public class DBAppender extends DBAppenderBase<ILoggingEvent> {
   protected String insertExceptionSQL;
   protected String insertSQL;
   protected static final Method GET_GENERATED_KEYS_METHOD;
+  protected String origin;
 
   private DBNameResolver dbNameResolver;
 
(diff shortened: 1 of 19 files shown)


logback-classic/src/main/java/ch/qos/logback/classic/db/DBAppender.java
```diff
@@ -62,7 +63,8 @@
   static final int  CALLER_CLASS_INDEX = 12;
   static final int  CALLER_METHOD_INDEX = 13;
   static final int  CALLER_LINE_INDEX = 14;
-  static final int  EVENT_ID_INDEX  = 15;
+  static final int  ORIGIN = 15;
```
tony19: Codacy Issue found: [Use explicit scoping instead of the default package private level](https://www.codacy.com/app/tony19/logback/file/2376070365/issues/source?bid=3222618&fileBranchId=3222712#l66)

logback-classic/src/main/java/ch/qos/logback/classic/db/DBAppender.java
```diff
@@ -62,7 +63,8 @@
...
+  static final int  EVENT_ID_INDEX  = 16;
```
tony19: Codacy Issue found: [Use explicit scoping instead of the default package private level](https://www.codacy.com/app/tony19/logback/file/2376070365/issues/source?bid=3222618&fileBranchId=3222712#l67)

PR #168, opened 2013-12-30
LOGBACK-945: extended ServerSocketReceiver and AbstractServerSocketAppender have their respective own thread pools for managing client connections

SierraGolf:
Previously the `ServerSocketReceiver` and the `AbstractServerSocketAppender` and their subclasses were using the context's thread pool for client connections which was effectively limiting the amount of possible client connections to a little less than the value of the `CoreConstants.MAX_POOL_SIZE` which is 32.

This pull requests extends the above mentioned classes so that they have their own thread pools dedicated for client connections.

The pools can be configured through the following new configuration properties:
- corePoolSize (default: CoreConstants.CORE_POOL_SIZE)
- maxPoolSize (default: CoreConstants.MAX_POOL_SIZE)

Questions:
1. Should I also extend the documentation with this pull request?
2. Should I implement some integration test in addition to the supplied unit tests to verify functional aspects of the new features?

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/CoreConstants.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/CoreConstants.java
@@ -13,16 +13,12 @@
  */
 package ch.qos.logback.core;
 
-import ch.qos.logback.core.util.EnvUtil;
-
 public class CoreConstants {
 
   /**
    * Number of idle threads to retain in a context's executor service.
    */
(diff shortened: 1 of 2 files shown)


SierraGolf:
@tony19 / @ceki  any chance this is merged any time soon?

melezov:
We would love for this patch to be merged.  
We're hitting the wall of 32 threads and cannot use the vanilla jar in production.

PR #181, opened 2014-03-27
Use ServiceLoader pattern to control logback bootstrap process

agentgt:
Currently there is no way to control the static initialization of logback such that you can do your own configuration of Logback with out hijacking an existing configuration. This change allows you to use the common [ServiceLoader ](http://docs.oracle.com/javase/6/docs/api/java/util/ServiceLoader.html) pattern from JDK6 to load a new SPI called `Configurator` that will allow custom initialization of the root appender similar to what BasicConfigurator is currently doing. I have Backported the `ServiceLoader` as its only available in JDK6 and above.

More detail is on my SO Post: http://stackoverflow.com/questions/22335441/configure-logback-to-defer-to-java-configuration-aka-plain-java-configuration-of


ittaiz:
ping?

PR #204, opened 2014-06-03
Proposal: Code Enforcement

hazendaz:
Updated codeStyle.xml to version 12

Added java-formatter-plugin which will enforce code style formatting on
java code.  It is turned off as it affects code beyond what I want to
touch.  Proposal is that if this is accepted and used, a one time
cleanup should be done by project owners to their liking as there may be
inconsistencies in the codeStyle.xml as compared to the actual code
base.  Rather than introducing that level of changes, I felt this would
be better done in that fashion.


hazendaz:
closing this since it isn't being accepted.

PR #207, opened 2014-06-16
Fix DateConverter to conform w/ISO8601 (LOGBACK-262)

tony19:
ISO8601 requires the letter "T" as the delimiter between the date and
time fields, but we were incorrectly using a space.

### DIFF
--- a/logback-classic/pom.xml
+++ b/logback-classic/pom.xml
@@ -142,6 +142,12 @@
       <version>2.0.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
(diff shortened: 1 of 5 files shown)


tony19:
The patch isn't really needed. The 'T' is optional, so we're already meeting the spec; and [most] users are fine with specifying custom patterns that include the 'T' if needed.

ceki:
Issue solved in commit 1b7fe9498d491e367d. Closing

See also documentation at https://logback.qos.ch/manual/layouts.html#date

bric3:
Then you

PR #217, opened 2014-07-18
New feature for the TimeBasedRollingPolicy: option to specify time zone ...

lilyevsky:
...and roll time.

Time zone: tag "timeZone", format - the usual Java time zone name.
Roll time: tag "rollTime", format "hh:mm:ss".

Example:

............................
<rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
            <timeZone>America/New_York</timeZone>
            <rollTime>13:15:00</rollTime>
      ...................................

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/rolling/DefaultTimeBasedFileNamingAndTriggeringPolicy.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/rolling/DefaultTimeBasedFileNamingAndTriggeringPolicy.java
@@ -39,12 +39,7 @@ public class DefaultTimeBasedFileNamingAndTriggeringPolicy<E> extends
   public boolean isTriggeringEvent(File activeFile, final E event) {
     long time = getCurrentTime();
     if (time >= nextCheck) {
-      Date dateOfElapsedPeriod = dateInCurrentPeriod;
-      if(dateOfElapsedPeriod != null) {
-          long tm = dateOfElapsedPeriod.getTime();
-          dateOfElapsedPeriod = new Date(tm + timeZone.getOffset(tm) - 
-                  TimeZone.getDefault().getOffset(tm));
-      }
+      Date dateOfElapsedPeriod = rc.getAdjustedTime(dateInCurrentPeriod);
       addInfo("Elapsed period: "+dateOfElapsedPeriod);
       elapsedPeriodsFileName = tbrp.fileNamePatternWCS
           .convert(dateOfElapsedPeriod);
--- a/logback-core/src/main/java/ch/qos/logback/core/rolling/DefaultTimeBasedFileNamingAndTriggeringPolicy.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/rolling/DefaultTimeBasedFileNamingAndTriggeringPolicy.java
@@ -17,6 +17,7 @@ import java.io.File;
 import java.util.Date;
 
 import ch.qos.logback.core.rolling.helper.TimeBasedArchiveRemover;
+import java.util.TimeZone;
 
 /**
  * 
@@ -39,6 +40,11 @@ public class DefaultTimeBasedFileNamingAndTriggeringPolicy<E> extends
     long time = getCurrentTime();
     if (time >= nextCheck) {
       Date dateOfElapsedPeriod = dateInCurrentPeriod;
+      if(dateOfElapsedPeriod != null) {
+          long tm = dateOfElapsedPeriod.getTime();
+          dateOfElapsedPeriod = new Date(tm + timeZone.getOffset(tm) - 
+                  TimeZone.getDefault().getOffset(tm));
+      }
       addInfo("Elapsed period: "+dateOfElapsedPeriod);
       elapsedPeriodsFileName = tbrp.fileNamePatternWCS
           .convert(dateOfElapsedPeriod);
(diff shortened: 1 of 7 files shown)


tony19:
Please review our [contributing guidelines](https://github.com/qos-ch/logback/blob/master/CONTRIBUTING.md). Your patches deviate from our set formatting style. Also, please squash your bug fixes into the offending patch (we can't accept fixes to bugs that originate from the same pull request).

The approval status of a pull request is always communicated in the pull request itself. If we reject your pull request, we would comment here and close it accordingly. If accepted, your pull request is merged. GitHub indicates whether the pull request could be automatically merged or whether a merge conflict exists, in which case you'd be responsible for rebasing your patch.

I don't quite understand the general usefulness of "daysOffset". Please provide a use case for this feature.

I don't agree with the proposed `timezone` and `rollTime` tags because they don't match how we currently handle time-related attributes of the rollover period (parsed from the filename pattern), and there's no compelling reason to change.

tinybox:
I would like to see that the log files could roll on a different time and day than what is hard coded. Our products have a lot of stuff going on at precisely 00:00 and that's when the logs are rolled. So we have to do every search in two log files.

That's a scenario for the feature. I don't think anyone is married to a specific way of solving it :)

tony19:
@tinybox, you're describing a use case for `rollTime`, which is clear to me. I'm questioning ["daysOffset"](https://github.com/lilyevsky/logback/commit/6174c14ec7c37f14886969384ea75a01707bf309), which IMO is not a worthwhile feature.

tinybox:
@tony19, I agree that everything should be done consistently. No new tags if it can be built in to the filename rolling syntax.

lilyevsky:
I appreciate everybody's comments, and I agree with most of them. Please, let's discuss in a constructive way, and let's solve this issue, it will be useful for many users.
As it says in the manual, "The rollover period is inferred from the value of fileNamePattern", and this is exactly where the problem comes from. When my log is archived, I may want its name to reflect the exact time when it happened, up to a minute or even second precision. But it does not mean that I want to roll every minute. Therefore, at least one additional parameter is necessary (unless we encrypt this additional information in the existing one, with some funny syntax - I believe, that would be not very convenient).
On the other hand, I introduced three new parameters, and I agree this is too many. Let's do some brainstorming here, maybe we can find a way which most users would like, hopefully with just one additional parameter using some clear syntax.
But let me reiterate the set of issues I am trying to solve. As I noted above, I believe that the rolling period optionally should be specified in a separate parameter (obviously we keep the default current behavior). In this parameter, we should be able to define the following:

- Time zone: I hope nobody would argue against this; very important for global apps
- Period length: day, hour, etc.
- When each period starts. Currently it is always the midnight, or top of the hour, etc., which is not what might be required by the application.

Please post you suggestions.

tony19:
Ok, I think I now see the value in separating the rollover period setting from the filename pattern, especially in cases where one would like to have regularly scheduled rollovers but with a file-naming scheme that didn't include the date (e.g., a simple incrementing counter).

I still don't understand the usefulness of `daysOffset` even with your JIRA comment below. Why would this be needed?
<blockquote>
I realized that the number of days to shift will be very useful, even for the daily rollover, so I added it, and submitted to my branch. It should be included in the pull. the new parameter is called "daysOffset", integer type (can be negative).
For daily rollover, this is how I use it myself. If daysOffset is zero (default), then the log before rollTime belongs to yesterday, and today starts after rollover. If I set it to 1, then the log before rollTime is today, and after rollTime it is tomorrow.
</blockquote>

bric3:
Hi all, I'm interested in this kind of feature as well.

However there's another issue with current implementation. The rollover only happen when there's a event sent `isTriggeringEvent` to the policy. Which may not happen until there's another log, depending on the frequency of some log events this can cause delays in the actual rollover.

This PR does seem to enable the rollover at a more precise _pace_ but doesn't seem to enforce the rollover at such defined time. I'm not sure how to do that without some sort of timer.

logback-core/src/main/java/ch/qos/logback/core/rolling/DefaultTimeBasedFileNamingAndTriggeringPolicy.java
```diff
@@ -17,6 +17,7 @@
 import java.util.Date;
 
 import ch.qos.logback.core.rolling.helper.TimeBasedArchiveRemover;
+import java.util.TimeZone;
```
tony19: Codacy Issue found: [Avoid unused imports such as 'java.util.TimeZone'](https://www.codacy.com/app/tony19/logback/file/2371918722/issues/source?bid=3222618&fileBranchId=3222699#l20)

logback-core/src/main/java/ch/qos/logback/core/rolling/helper/RollingCalendar.java
```diff
@@ -237,7 +237,18 @@ public Date getRelativeDate(Date now, int periods) {
...
   }
 
-  public Date getNextTriggeringDate(Date now) {
-    return getRelativeDate(now, 1);
-  }
+    public Date getNextTriggeringDate(Date now) {
+        return getRelativeDate(now, 1);
+    }
+
+    public Date getAdjustedTime(Date d) {
```
tony19: Codacy Issue found: [Avoid reassigning parameters such as 'd'](https://www.codacy.com/app/tony19/logback/file/2371918749/issues/source?bid=3222618&fileBranchId=3222699#l244)

lilyevsky:
This issue became irrelevant, so I am closing it.
First reason - you can basically achieve the desired affect by using auxiliary format elements and adding time zone to the format.
Second reason - I am using log4j2 now.

PR #225, opened 2014-10-12
LOGBACK-910: Set the interrupt flag when we catch InterruptException to ...

henrikno:
...avoid swallowing it.
We are currently using a library which uses the interrupt flag to abort the current job. If we log at the wrong time we currently lose that notification, and carry on like nothing happened. This can lead to critical errors in our case.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/AsyncAppenderBase.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/AsyncAppenderBase.java
@@ -134,6 +134,7 @@ public class AsyncAppenderBase<E> extends UnsynchronizedAppenderBase<E> implemen
       
     } catch (InterruptedException e) {
       addError("Failed to join worker thread. " + blockingQueue.size() + " queued events may be discarded.", e);
+      Thread.currentThread().interrupt();
     }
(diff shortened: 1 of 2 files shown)


henrikno:
<p>Fixed a bug in <code>AsyncAppenderBase<code> where the interrupt flag was not restored on <code>InterruptedException</code>.</p>

kay:
Just to thrown in my support for this PR in the next release - I've came across this issue today with the same scenario. It means we cannot use the AsyncAppender with any interruptible code that logs. I was just about to raise my own PR with the same fix. So thanks @henrikno !

ceki:
I'll be reviewing this issue in the coming hours.

ceki:
Many thanks for opening this pull request. Problem fixed in https://github.com/qos-ch/logback/commit/df0e5f936755cc866ff65ff9488dcfc1b 

The aforementioned commit is very close to your proposal. The only difference is that the worker thread, which is private to AsyncAppender, still consumes the `InterruptedException` without calling `Thread.currentThread().interrupt()`. I am assuming that only AsyncAppnder can interrupt its worker thread.

PR #226, opened 2014-11-04
Fix for LOGBACK-840

oyvindn:
Should not log warn for SizeAndTimeBasedArchiveRemover when the periodsElapsed is less than 1, as it is expected behavior. If the log is set to roll once per day, but is rolled because of size in between, the periodsElapsed will be less than 1.


wimdeblauwe:
I noticed this warning "Unexpected periodsElapsed value 0" in my log as well. Is this going to get fixed in the next version?

PR #239, opened 2015-01-19
Change default PatternLayout to %ex (LOGBACK-966)

pedro132:
Set the default PatternLayout to use exception tracing without calculating packaging data when no Converter capable of handling exceptions is specified ([LOGBACK-966](http://jira.qos.ch/browse/LOGBACK-966)).

I changed the ExtendedThrowableProxyConverterTest because it was failing as it was, but I don't know if further changes are needed to it or ThrowableProxyConverterTest due to the fact that the default converter was altered. Please tell me if anything is missing.

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/pattern/EnsureExceptionHandling.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/pattern/EnsureExceptionHandling.java
@@ -24,7 +24,7 @@ public class EnsureExceptionHandling implements
   /**
    * This implementation checks if any of the converters in the chain handles
    * exceptions. If not, then this method adds a
-   * {@link ExtendedThrowableProxyConverter} instance to the end of the chain.
+   * {@link ThrowableProxyConverter} instance to the end of the chain.
    * <p>
    * This allows appenders using this layout to output exception information
    * event if the user forgets to add %ex to the pattern. Note that the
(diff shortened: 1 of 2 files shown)


pedro132:
It looks like the build failure was unrelated to my commit, as the failing tests were on logback-core and I only changed code from logback-classic. I also re-run the tests on my machine and the passed. By the logs the cause looks a timed out test.

tony19:
Thanks for the contribution, @pedro132.  Please add a comment here with release notes (see rule P7 of [CONTRIBUTING.md](https://github.com/qos-ch/logback/blob/master/CONTRIBUTING.md)).

pedro132:
My pleasure.

```html
When using the default PatternLayout, if no Converter capable of handling exceptions
is specified, use exception tracing without calculating packaging data (see 
<a href="http://jira.qos.ch/browse/LOGBACK-966">LOGBACK-966</a> for details).
```

tony19:
The expensive packaging data calculation actually occurs when the `LoggingEvent` is created, and `ExtendedThrowableProxyConverter` simply formats the already calculated data. To address the problem, we've disabled the packaging calculation by default (8141f88edb95567c2b42bbc73aa81e9eeb399169). When no packaging data is available, the `ExtendedThrowableProxyConverter` is effectively the same as `ThrowableProxyConverter`, so this PR is no longer needed.

pedro132:
All right then.

PR #241, opened 2015-02-03
LOGBACK-1046 fix suggestion

nadavwe:
as described in Jira, releasing the lock on a closed channel results in an Exception that renders the appender non-started for the rest of the application's life.
checking that the channel is still open before releasing it mitigates this problem.


nadavwe:
changed commit after @tony19 suggestion.

tony19:
Thanks for the contribution, @nadavwe.  Please add a comment here with release notes (see rule P7 of [CONTRIBUTING.md](https://github.com/qos-ch/logback/blob/master/CONTRIBUTING.md)).

nadavwe:
file appenders with prudent mode will now recover after an IO error, instead of becoming non-started.

PR #248, opened 2015-03-07
Add XML support to en/disable stacktrace package detail

tony19:
Changed the default enable status of packaging info in stacktraces
to be disabled to avoid the expensive lookup unless requested.

Added an XML attribute "packageTrace" to <configuration> in order to
allow enabling or disabling the packaging info in stacktraces. When
the attribute is omitted, the default setting is assumed (disabled).

To enable package detail in logback.xml:

```xml
  <configuration packageTrace="true">
    ...
  </configuration>
```

LOGBACK-730
LOGBACK-966


pierre:
:+1:

Could this land in 1.1.3?

PR #249, opened 2015-03-15
LOGBACK-488 Enable 'hostname' variable to be accessible from nested scope when using Groovy DSL

yihtserns:
Logback's Groovy doc [mentioned](http://logback.qos.ch/manual/groovy.html#hostname) that:
> ...due to scoping rules that the authors cannot fully explain, the 'hostname' variable is available only at the topmost scope but not in nested scopes.

The reason for this is because while `ComponentDelegate.propertyMissing(String, def)` was meant (I'm guessing) to _only_ resolve missing property **setter**, Groovy _also_ used it to resolve missing property **getter**.

The latter would silently fail, making Groovy thinks everything is OK.

### Documentation
Rule P8 of [CONTRIBUTING.md](https://github.com/qos-ch/logback/blob/master/CONTRIBUTING.md) says (the above) documentation must be updated, but I'm not sure how it should be written...

### Release note
```
<p>'hostname' variable is now accessible from nested scope when using Groovy DSL. (<a href="http://jira.qos.ch/browse/LOGBACK-488">LOGBACK-488</a>)</p>
```

### DIFF
--- a/logback-classic/src/main/groovy/ch/qos/logback/classic/gaffer/ComponentDelegate.groovy
+++ b/logback-classic/src/main/groovy/ch/qos/logback/classic/gaffer/ComponentDelegate.groovy
@@ -85,7 +85,7 @@ class ComponentDelegate extends ContextAwareBase {
     }
   }
 
-  void propertyMissing(String name, def value) {
+  void set(String name, def value) {
     NestingType nestingType = PropertyUtil.nestingType(component, name);
     if (nestingType == NestingType.NA) {
       addError("${getLabelFistLetterInUpperCase()} ${getComponentName()} of type [${component.getClass().canonicalName}] has no appplicable [${name}] property ")
--- a/logback-classic/src/test/groovy/ch/qos/logback/classic/gaffer/GafferConfiguratorTest.groovy
+++ b/logback-classic/src/test/groovy/ch/qos/logback/classic/gaffer/GafferConfiguratorTest.groovy
@@ -161,6 +161,16 @@ class GafferConfiguratorTest {
         assertEquals("HELLO %m%n", ca.encoder.getLayout().pattern)
     }
 
+    @Test
+    public void hostnameCascading() {
+        File file = new File(ClassicTestConstants.GAFFER_INPUT_PREFIX + "hostnameCascading.groovy")
+        String dslText = file.text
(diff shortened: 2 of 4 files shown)


tony19:
Since your patch fixes the problem indicated in the docs, the problem should be removed from the docs.

ceki:
>The reason for this is because while ComponentDelegate.propertyMissing(String, def) was meant (I'm > guessing) to only resolve missing property setter, Groovy also used it to resolve missing property 
> getter.

I am not sure I follow. What's this set(String name, def value) method and how/why does it apply in this context?

yihtserns:
@tony19 Shouldn't the doc say something like "it is not a problem since version X", instead of being totally removed?

@ceki `set(String name, value)` is something called ["generic set method"](https://github.com/groovy/groovy-core/blob/GROOVY_1_8_0/src/main/org/codehaus/groovy/runtime/MetaClassHelper.java#L773).  Generic get/set method is one level above `propertyMissing(...)`.

And yeah, I've also been searching high and low for the doc, but found nothing.  I don't even remember how I knew about this.  The closest I found is just this ["generic get method" mentioned by Kohsuke](https://jira.codehaus.org/browse/GROOVY-3867).

If you're not comfortable with this, another way to fix would probably be to declare a `propertyMissing(String)` that simply exists to throw `MissingPropertyException` for missing getters.

yihtserns:
Found ["generic get/set method" mentioned in Groovy's mailing list](http://groovy.329449.n5.nabble.com/template/NamlServlet.jtp?macro=search_page&node=329449&query=%22generic+get%2Fset%22+OR+%22get+vs+getProperty%22&days=0).

Decided to [ask whether it is safe to use](http://groovy.329449.n5.nabble.com/Is-it-safe-to-use-generic-get-set-method-td5723173.html), and also whether [missingProperty(String, Object) behaviour is bug](http://groovy.329449.n5.nabble.com/missingProperty-String-Object-handles-both-missing-setter-AND-getter-td5723174.html).

yihtserns:
This issue is actually caused by a bug in Groovy ([GROOVY-7723](https://issues.apache.org/jira/browse/GROOVY-7723)).

It is fixed in (the upcoming) Groovy version 2.4.7.

PR #251, opened 2015-04-08
replaces <span> with <b> so the viewAsGroovy Link works again

klausbayrhammer:
With this commit the viewAsGroovy link works again. Although I'm not sure the generated groovy config is correct because the debug() statement is missing.


klausbayrhammer:
Thanks for the hint. Fixing the regex in the ```asGroovy``` function is a way better solution.

tony19:
Thanks for the update. Can you squash these commits (and provide a quick blurb here for release notes)?

PR #254, opened 2015-04-14
Optimization in LogbackMDCAdapter.get(String)

panchenko:
This function returns only a single String value, so recording that last operation was read is not necessary.


panchenko:
Fixed the LogbackMDCAdapterTest.sequenceWithGet() test.

ceki:
The code in question is on the tricky side and would have merited a little more discussion. The fact that a failing test case was modified to accommodate the change should have raised suspicion.  As it happens and to my best judgement, the change it is correct.

PR #260, opened 2015-05-30
LOGBACK-1075 and LOGBACK-1076

Janario:
Changes to work on modular application.
In my case Wildfly, logback as a module

More details in the issues LOGBACK-1075 and LOGBACK-1076


tony19:
This looks like a duplicate of #255.

Janario:
@tony19,
Yes it is quite related to #255 
But as I've seen none was integrated and still a problem.

In my pull request I've done changes to try as a fallback (not global) and may not affect other users once they should still work as it is already in current logback. To users that fails it will try from other CL and works if it found resources

- resources LOGBACK-1075 <property resource="my-file.properties"/>
- application appender LOGBACK-1076
- conditional with janino LOGBACK-1077

mkurz:
Since #255 never made it in coul we give that pull request another chance?

nafg:
Pretty please?

PR #268, opened 2015-07-16
LOGBACK-1027 - correctly handle circular suppressed and cause exceptions

elindsey:
Previously logback would infinitely recurse when encountering a circular reference in the cause or suppressed exception chain.

Now it will detect circular references and output an annotated stack trace (see commit message for a formatted example).

Note that this differs slightly from normal Throwable.printStackTrace() behavior in that it preserves context around where the circular reference was found (ie. will still print out Suppressed or Caused by).

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/spi/ThrowableProxy.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/spi/ThrowableProxy.java
@@ -17,6 +17,9 @@ import ch.qos.logback.core.CoreConstants;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Set;
 
 public class ThrowableProxy implements IThrowableProxy {
 
@@ -28,7 +31,7 @@ public class ThrowableProxy implements IThrowableProxy {
   // package-private because of ThrowableProxyUtil
   int commonFrames;
   private ThrowableProxy cause;
-  private ThrowableProxy[] suppressed=NO_SUPPRESSED;
+  private ThrowableProxy[] suppressed = NO_SUPPRESSED;
 
   private transient PackagingDataCalculator packagingDataCalculator;
   private boolean calculatedPackageData = false;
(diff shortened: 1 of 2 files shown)


SvetlinZarev:
Hi,

Please add a test case to check if your fix works for this case:


    public class Sample {

		try {
			try (CloseMe closeMe = new CloseMe()) {
				throw new RuntimeException("From TRY");
			}
		} catch (Exception exception) {
			System.out.println("Exception: " + exception);
			System.out.println("Suppressed: " + exception.getSuppressed().length);
			System.out.println("Suppressed: " + exception.getSuppressed()[0]);
			System.out.println("Suppressed cause: " + exception.getSuppressed()[0].getCause());

			System.out.println("Setting the cause of the suppressed exception"); 
			exception.getSuppressed()[0].initCause(exception);
			System.out.println("Suppressed cause: " + exception.getSuppressed()[0].getCause());
		}
	}
}

class CloseMe implements AutoCloseable {
	@Override
	public void close() throws RuntimeException {
		throw new RuntimeException("On Close"); // Will be suppressed by the one
												// from the try body
	}
}

dfjones:
@SvetlinZarev 

Looks like source and target compatibility with Java 6 is declared for this project, but the try-with-resources and suppressed exception behavior you are asking me to test is only available starting in Java 7. Is there some place to write test cases against Java 7 behavior?

Assuming the answer is no, I believe the following tests the situation you are asking about, only with manually creating the chain of suppressed exception with a circular cause. Please take a look and let me know if this is sufficient. If so, I will update the PR with this test case.

```java
  @Test
  public void testSuppressedCircularCause() {
    Exception ex = new Exception("Top");
    Exception suppressed = new Exception("Suppressed");

    ex.addSuppressed(suppressed);
    assertEquals(1, ex.getSuppressed().length);
    Throwable actualSuppressed = ex.getSuppressed()[0];
    assertEquals(suppressed, actualSuppressed);
    assertNull(actualSuppressed.getCause());

    actualSuppressed.initCause(ex);
    verifyContains(ex, "Caused by: CIRCULAR REFERENCE:java.lang.Exception: Top");
  }
```

dfjones:
Any word on this PR? I'm happy to make any changes/add additional tests to get this into a state that is ready for merge.

PR #269, opened 2015-08-05
LOGBACK-1052 Have RequestLogImpl implement Jetty's LifeCycle interface

kylesm:
Have RequestLogImpl implement Jetty's LifeCycle interface directly. This
allows logback access to work with Jetty 9.3.0M1 and newer, where the
Jetty API was changed in an incompatible way.

### DIFF
--- a/logback-access/pom.xml
+++ b/logback-access/pom.xml
@@ -46,6 +46,12 @@
       <optional>true</optional>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-util</artifactId>
+      <scope>compile</scope>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.codehaus.janino</groupId>
       <artifactId>janino</artifactId>
       <scope>compile</scope>
--- a/logback-access/src/main/java/ch/qos/logback/access/jetty/RequestLogImpl.java
+++ b/logback-access/src/main/java/ch/qos/logback/access/jetty/RequestLogImpl.java
@@ -26,6 +26,7 @@ import ch.qos.logback.core.util.StatusPrinter;
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.RequestLog;
 import org.eclipse.jetty.server.Response;
+import org.eclipse.jetty.util.component.LifeCycle;
 
 import ch.qos.logback.access.joran.JoranConfigurator;
 import ch.qos.logback.access.spi.AccessEvent;
@@ -112,7 +113,7 @@ import ch.qos.logback.core.util.OptionHelper;
  * @author S&eacute;bastien Pennec
  */
 public class RequestLogImpl extends ContextBase implements RequestLog,
-        AppenderAttachable<IAccessEvent>, FilterAttachable<IAccessEvent> {
+        AppenderAttachable<IAccessEvent>, FilterAttachable<IAccessEvent>, LifeCycle {
 
   public final static String DEFAULT_CONFIG_FILE = "etc" + File.separatorChar
           + "logback-access.xml";
--- a/pom.xml
+++ b/pom.xml
@@ -184,6 +184,11 @@
         <version>${jetty.version}</version>
       </dependency>
       <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-util</artifactId>
+        <version>${jetty.version}</version>
+      </dependency>
+      <dependency>
         <groupId>org.apache.geronimo.specs</groupId>
         <artifactId>geronimo-jms_1.1_spec</artifactId>
         <version>1.0</version>


kylesm:
<p>Have <code>RequestLogImpl</code> implement Jetty's <code>LifeCycle</code> interface directly. This allows logback access to work with Jetty 9.3.0M1 and newer, where the Jetty API was changed in an incompatible way (<a href="http://jira.qos.ch/browse/LOGBACK-1052">LOGBACK-1052</a>).
</p>

blop:
+1 - Jetty 9.3 is released now and this needs to be fixed :)

jdillon:
+1 to get this fixed, logback-access is useless asis in Jetty 9.3+ w/o this change.

BTW I tested locally on master and I only had to add Jetty LifeCycle to the impl, no other changes.

fzakaria:
Any update on this?

tpounds:
FWIW, you can workaround this issue by extending the base class and adding the missing interface as a stopgap.

```Java
package com.example.ch.qos.logback.access.jetty;

import ch.qos.logback.access.jetty.RequestLogImpl;
import org.eclipse.jetty.util.component.LifeCycle;

public class RequestLogImpl_Fix_LOGBACK_1052 extends RequestLogImpl implements LifeCycle {
}
```

The Jetty config would just need to be updated to use the wrapper class.

```XML
<Ref id="RequestLogHandler">
    <Set name="requestLog">
        <New id="requestLogImpl" class="com.epicgames.ch.qos.logback.access.jetty.RequestLogImpl_Fix_LOGBACK_1052"/>
    </Set>
</Ref>
```

robmoore:
If the fix doesn't seem imminent, I'd like to suggest the docs be updated to reflect this situation. I spent an unfortunate amount of time trying to work out why it's not working properly and would like to save others the pain.

robmoore:
@kylesm Do you happen to know if this fix will be backwards compatible with older versions of Jetty? Just curious if this interface was introduced in the 9.x line.

ksmith14:
@robmoore the LifeCycle interface has been around since at least Jetty 6.1.2. A couple of methods were added to it in Jetty 7 but RequestLogImpl has stubbed out versions of them (add/removeLifeCycleListener), so it shouldn't alter the minimum supported version of Jetty.

robmoore:
@kylesm Thanks, Kyle. I was thinking that perhaps the issue of backwards compatibility was what might be holding this fix up. Good to know it's not an issue.

joakime:
The `RequestLog` has transitioned from being a traditional `Handler` in the Jetty sense, to being a specialized implementation that the `HttpChannel` uses.

This was done to address many a bug with async request logging, along with request logging of bad requests (ones that cannot be sent to a context).

Since you are using slf4j & logback, you no longer need logback-access in Jetty 9.3+

Use the [`org.eclipse.jetty.server.Slf4jRequestLog`](http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/server/Slf4jRequestLog.html) implementation and just configure your normal logback to to route [the named logging](http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/server/Slf4jRequestLog.html#setLoggerName-java.lang.String-) to the log appender of your choice.

ChrisMacpherson-awin:
@joakime That does work, but logback-access has specific access log PatternLayout that is really useful in order to create the log format. 

I cant see how to create a similar access log format using the pattern facilities in Logback itself without having to write a load of code. 

Do you know of a way?

joakime:
@ChrisMacpherson-awin seeings as the Jetty `Slf4jRequestLog` is an implementation of the `AbstractNCSARequestLog` it it satisfying the NCSA output format correctly.

Altering it with a customizable pattern would no longer make it NCSA format compatible.

The logback-access project could implement their own [`RequestLog`](http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/server/RequestLog.html) and use it as an implementation in Jetty, but it still wouldn't be part of the Jetty `LifeCycle`. (the original purpose of this particular pull request)

ChrisMacpherson-awin:
@joakime Okay maybe I just didn't understand the hook up correctly. In logback it needs an encoder in the appender that you use in the named logger. I couldn't work out how to configure it to output the request log data directly without using a pattern. Do you know of any examples anywhere. I couldn't find any. 

Anyway I've probably digressed from the purpose of this pull request

kulinski:
Is there an update on why this PR couldn't be incorporated?  We wasted many hours tracking down this issue for what looks like a simple fix that could have been implemented years ago.  The current documentation for Jetty 9.3+ and Logback doesn't indicate that `logback-access` is now incompatible.

We also are using logback custom layouts so switching to `Slf4jRequestLog` isn't a proper resolution for us.  We're currently using the workaround suggested by @tpounds

davidyscott:
Bump?

mkristian:
we also using @tpounds workaround - rebase/merge this PR please !

xak2000:
What wrong with this PR (I mean why it isn't accepted)? I'm one of those folks that spent several hours of debugging to understand why the [documented](https://logback.qos.ch/access.html#jetty) behaviour of `logback-access` library isn't working: the `start()` method isn't called by Jetty because of this `Lifecycle` absense. And without debugging and deep knowledge of Jetty internals, it looks like it just silently doesn't work for an unknown reason.

This PR is almost 4 years old and all this time people like me spend time for no reason. If several years ago Jetty 9.3 can be seen as something new, now it is very actual version, if not old (I use 9.4.7 and current version is 9.4.15). So with time passing this PR's demand will only increase.

joakime:
Note that Jetty 9.4.15+ has a CustomRequestLog that accepts formatting.

Example usage (with slf4j)

``` java
        Slf4jRequestLogWriter slf4jRequestLogWriter = new Slf4jRequestLogWriter();
        slf4jRequestLogWriter.setLoggerName("my.requestlog"); // the slf4j named logger to output to
        // The format of the output. (See javadoc for CustomRequestLog for details)
        String requestLogFormat = "%{client}a - %u %t \"%r\" %s %O \"%{Referer}i\" \"%{User-Agent}i\"";
        CustomRequestLog requestLog = new CustomRequestLog(slf4jRequestLogWriter, requestLogFormat);
        server.setRequestLog(requestLog);
```

xak2000:
Thanks @joakime ! `CustomRequestLog` is good if it fulfill your requirements.

But unfortunately, it doesn't have parity in functionality with `logback-access`.

For example, what I need:

- Full request/response headers (not just one header with specific name, but all).
- Full request/response body (I understand that this requires buffering but when you need debug functionality and load isn't too high - it is acceptable). `logback-access` supports this out of the box with `TeeFilter` and you can always disable that filter if debug is off.
- Ability to filter by request URL (e.g. to prevent loadbalancer's healthcheck spamming). See `ch.qos.logback.core.filter.EvaluatorFilter` and `ch.qos.logback.access.net.URLEvaluator`.

The main difference between `logback-access` and other approaches like `CustomRequestLog` is that with `logback-access` you have access to current http request/response from the logging event (see `ch.qos.logback.access.spi.AccessEvent`), while at the same time you can use some functionality of logback (filtering). On the other hand, `CustomRequestLog` passes just an opaque string to the Slf4j logger. You can't use this string to filtering on logback side. Mabye this can be achievable by using some MDC keys containing request/response info.

PR #272, opened 2015-09-08
LOGBACK-1087 fix (with test)

UndefinedBehaviour:
Uses a (shallow) copy of the statusListenerList to invoke addStatusEvent, so there are no concurrency issues if the original list gets modified at the same time (a somehow convoluted but legitimate usage).

### DIFF
--- a/logback-core/src/test/java/ch/qos/logback/core/BasicStatusManagerTest.java
+++ b/logback-core/src/test/java/ch/qos/logback/core/BasicStatusManagerTest.java
@@ -87,5 +87,14 @@ public class BasicStatusManagerTest {
     }
   }
 
-
+  @Test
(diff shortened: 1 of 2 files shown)


logback-core/src/test/java/ch/qos/logback/core/BasicStatusManagerTest.java
```diff
@@ -87,5 +87,14 @@ public void duplicateInstallationsOfOnConsoleListener() {
     }
   }
 
-
+  @Test
+  public void testFireStatusAddEventConcurrentModificationException() {
```
tony19: Codacy Issue found: [JUnit tests should include assert() or fail()](https://www.codacy.com/app/tony19/logback/file/2371722102/issues/source?bid=3222618&fileBranchId=3222683#l91)

PR #185, opened 2014-04-03
Introduce AsyncAppender for logback-access: LOGBACK-827

kpavlov:
Introduced `ch.qos.logback.access.AsyncAppender` which accepts `IAccessEvents`. 
Standard AsyncAppender from logback-classic does not support IAccessEvents and fails with exception:

```
18:31:29,255 |-ERROR in ch.qos.logback.classic.AsyncAppender[access.file.async] - Appender [access.file.async] failed to append. java.lang.ClassCastException: ch.qos.logback.access.spi.AccessEvent cannot be cast to ch.qos.logback.classic.spi.ILoggingEvent
    at java.lang.ClassCastException: ch.qos.logback.access.spi.AccessEvent cannot be cast to ch.qos.logback.classic.spi.ILoggingEvent
    at  at ch.qos.logback.classic.AsyncAppender.preprocess(AsyncAppender.java:28)
```

Proposed fix for http://jira.qos.ch/browse/LOGBACK-827

Example `logback-access.xml`:

``` xml
<?xml version='1.0' encoding='utf-8'?>
<configuration>

    <!-- always a good activate OnConsoleStatusListener -->
    <statusListener class="ch.qos.logback.core.status.OnConsoleStatusListener"/>

    <!-- See for reference: http://logback.qos.ch/access.html#configuration -->
    <appender name="access.file" class="ch.qos.logback.core.rolling.RollingFileAppender">
        <file>${catalina.base}/logs/access.log</file>
        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
            <!-- rollover daily -->
            <fileNamePattern>${catalina.base}/logs/access.%d{yyyy-MM-dd}.%i.log.tar.gz</fileNamePattern>
            <timeBasedFileNamingAndTriggeringPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedFNATP">
                <!-- or whenever the file size reaches 100MB -->
                <maxFileSize>100KB</maxFileSize>
            </timeBasedFileNamingAndTriggeringPolicy>
        </rollingPolicy>
        <encoder>
            <pattern>combined</pattern>
        </encoder>
    </appender>

    <appender name="access.file.async" class="ch.qos.logback.access.AsyncAppender">
        <appender-ref ref="access.file"/>
    </appender>

    <appender-ref ref="access.file.async"/>
</configuration>
```

PR #277, opened 2015-10-14
[LOGBACK-827] restored solution https://github.com/qos-ch/logback/pul…

atais:
I have found an old solution by @kpavlov and tried to make it work for current version.

I have managed to move most of the changes from https://github.com/qos-ch/logback/pull/185

The only part left from `logback-access/src/test/java/ch/qos/logback/access/dummy/DummyRequest.java`
```java

        public Object getAttribute(String key) {
            if (key.equals("testKey")) {
                return "testKey";
            } else if (AccessConstants.LB_INPUT_BUFFER.equals(key)) {
                return DUMMY_CONTENT_BYTES;
            } else if (AccessConstants.LB_OUTPUT_BUFFER.equals(key)) {
                return DUMMY_RESPONSE_CONTENT_BYTES;
            } else {
                return null;
            }
        }
```

But i do not see the reason for this code. I quickly tested the solution with Tomcat 7 and it seems to work.

### DIFF
--- a/logback-access/src/main/java/ch/qos/logback/access/AsyncAppender.java
+++ b/logback-access/src/main/java/ch/qos/logback/access/AsyncAppender.java
@@ -1,0 +1,7 @@
+package ch.qos.logback.access;
+
+/**
+ * Created by msiatkowski on 14.10.15.
+ */
+public class AsyncAppender {
+}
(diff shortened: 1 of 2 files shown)


atais:
Well that is strange.. I did not get an error on my PC.

PR #280, opened 2015-10-20
LOGBACK-1092: Share GEventEvaluator with logback-access

mwhipple:
create a GEventEvaluatorBase in core which is extended in classic and access
http://jira.qos.ch/browse/LOGBACK-1092

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/boolex/GEventEvaluatorBase.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/boolex/GEventEvaluatorBase.java
@@ -1,0 +1,85 @@
+/**
+ * Logback: the reliable, generic, fast and flexible logging framework.
+ * Copyright (C) 1999-2015, QOS.ch. All rights reserved.
+ * <p>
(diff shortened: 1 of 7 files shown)


logback-core/src/main/java/ch/qos/logback/core/boolex/GEventEvaluatorBase.java
```diff
@@ -0,0 +1,83 @@
...
+ * under the terms of the GNU Lesser General Public License version 2.1
+ * as published by the Free Software Foundation.
+ */
+package ch.qos.logback.core.boolex;
+
+import ch.qos.logback.core.util.FileUtil;
+import groovy.lang.*;
+import org.codehaus.groovy.control.CompilationFailedException;
+
+public class GEventEvaluatorBase<E> extends EventEvaluatorBase<E> {
```
mwhipple: I'd think this and `IEvaluator` should possibly be `E extends DeferredProcessingAware` but I'll defer making the adjustment until there seems to be activity.

PR #282, opened 2015-11-03
LOGBACK-1094: Fix SSL configuration

romainmoreau:
Fix for LOGBACK-1094
More informations about the issue and the fix :
https://java.net/projects/javamail/forums/forum/topics/223631-Exception-sending-email-through-smtp-port-465


romainmoreau:
Since it's seems there is more activity on this project those days, would it be useful for me to fix the conflict there is on my commit? To make it short, are you planning to merge this pull request in a near future? Is there anything else I can do on it to help?

ceki:
Please create a Jira issue and mention this PR.

ceki:
BTW, logback uses javax.mail version 1.4 whereas your fix is for version 1.5.3? Is that correct?

romainmoreau:
That's right, this fix works with the version 1.5.3 and after. Don't you plan on moving to the 1.5 API? I suppose you are keeping the 1.4 version for backwards compatibility but it is really old.

PR #284, opened 2015-11-12
LOGBACK-1054 NullPointerException from RollingFileAppender

palek:
Simple fix and a test case for the root cause.


twothirdscyft:
Confirmed work around for NullPointException is using absolute paths in the XML configuration for 1.1.3.

```xml
  <appender name="file" class="ch.qos.logback.core.rolling.RollingFileAppender">
    <file>/path/to/logs/application.log</file>
    <rollingPolicy class="ch.qos.logback.core.rolling.FixedWindowRollingPolicy">
      <fileNamePattern>/path/to/logs/application.%i.log</fileNamePattern>
      <minIndex>1</minIndex>
      <maxIndex>2</maxIndex>
    </rollingPolicy>
    ...
```

PR #289, opened 2015-11-29
[LOGBACK-1114] Migrate from fest-assert to assertj

hazendaz:
fest-assert is a defunct library which was replaced by assertj.

migration notes here.

http://joel-costigliola.github.io/assertj/assertj-core-migrating-from-fest.html#fest-1.4

Using version 1.7.1 as that is last to support java 6.  Assertj 2.x is for java 7 building which is possible and still targetting java 6.  Assertj 3.x is for java 8 building which again is possible to still target java 6.  In order to not make such changes, will stick to version 1.7.1.

### DIFF
--- a/logback-classic/src/test/java/ch/qos/logback/classic/db/SQLBuilderTest.java
+++ b/logback-classic/src/test/java/ch/qos/logback/classic/db/SQLBuilderTest.java
@@ -13,7 +13,7 @@
  */
 package ch.qos.logback.classic.db;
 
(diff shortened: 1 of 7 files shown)


hazendaz:
@ceki can I get a status on this?  It's a rather simple change with little to no impact.

hazendaz:
@ceki Code has been rebased.  Any status on pulling this in?  Rather trivial change as it only applies to test side of the code and keeps it on the right path for this library change.

ceki:
Thank you.

PR #290, opened 2015-12-02
LOGBACK-898 never blocking async appender

tempredirect:
Allows the AsyncAppender to be configured to be never block on
a full queue.

This is taken from Jeff Wartes original patch:

http://jira.qos.ch/browse/LOGBACK-898

I submitted a signed CLA some months ago.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/AsyncAppenderBase.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/AsyncAppenderBase.java
@@ -51,6 +51,7 @@ public class AsyncAppenderBase<E> extends UnsynchronizedAppenderBase<E> implemen
 
   static final int UNDEFINED = -1;
   int discardingThreshold = UNDEFINED;
+  boolean neverBlock = false;
 
   Worker worker = new Worker();
   
@@ -152,12 +153,16 @@ public class AsyncAppenderBase<E> extends UnsynchronizedAppenderBase<E> implemen
   }
 
   private void put(E eventObject) {
-    try {
-      blockingQueue.put(eventObject);
-    } catch (InterruptedException e) {
-      // Interruption of current thread when in doAppend method should not be consumed
-      // by AsyncAppender
-      Thread.currentThread().interrupt();  	
+    if (neverBlock) {
+      blockingQueue.offer(eventObject);
+    } else {
+      try {
+        blockingQueue.put(eventObject);
+      } catch (InterruptedException e) {
+        // Interruption of current thread when in doAppend method should not be consumed
+        // by AsyncAppender
+        Thread.currentThread().interrupt();
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


stevenschlansker:
+1, we need this too.  Will test it out and report back our experience.

logback-core/src/main/java/ch/qos/logback/core/AsyncAppenderBase.java
```diff
@@ -152,12 +153,16 @@ private boolean isQueueBelowDiscardingThreshold() {
...
 
   private void put(E eventObject) {
-    try {
-      blockingQueue.put(eventObject);
-    } catch (InterruptedException e) {
-      // Interruption of current thread when in doAppend method should not be consumed
-      // by AsyncAppender
-      Thread.currentThread().interrupt();  	
+    if (neverBlock) {
+      blockingQueue.offer(eventObject);
```
stevenschlansker: Would it be safe to log a `WARN` status once (*not* a log message obviously, but a context status) when this happens?  Otherwise you would never know we lost your logs...

randomstatistic: At this point, if you're using AsyncAppender, you're already silently dropping logs at INFO level and below. 

If you have a mechanism to safely notify that we dropped something, that couldn't hurt, but I don't think this patch makes things any worse. You've already explicitly said (in your config) you prefer dropping your ERROR logs instead of blocking your application.

I do vaguely recall I thought about setting a flag and constructing and emitting a notification log message if the queue cleaned out again, but testing the edge boundary seemed tricky, since you might be flapping on the edge of capacity and make the problem worse.

tempredirect:
@ceki thank you

Pro100AlexHell:
This code have a bug
when discardingThreshold != 0, but 51 by default for example, then append() will return and discard, regardless of neverBlock true\false value

`protected void append(E eventObject) {
        if (isQueueBelowDiscardingThreshold() && isDiscardable(eventObject)) {
            return;
        }
        preprocess(eventObject);
        put(eventObject);
    }

    private boolean isQueueBelowDiscardingThreshold() {
        return (blockingQueue.remainingCapacity() < discardingThreshold);
    }

    private void put(E eventObject) {
        if (neverBlock) {
            blockingQueue.offer(eventObject);
        } else {
            putUninterruptibly(eventObject);
        }
    }`

you can correct this code
or correct the doc which says
https://logback.qos.ch/manual/appenders.html#asyncNeverBlock
>>
neverBlock 	boolean 	If false (the default) the appender will block on appending to a full queue rather than losing the message

By default, event queue is configured with a maximum capacity of 256 events. If the queue is filled up, then application threads are blocked from logging new events until the worker thread has had a chance to dispatch one or more events. When the queue is no longer at its maximum capacity, application threads are able to start logging events once again. Asynchronous logging therefore becomes pseudo-synchronous when the appender is operating at or near the capacity of its event buffer. This is not necessarily a bad thing. The appender is designed to allow the application to keep on running, albeit taking slightly more time to log events until the pressure on the appenders buffer eases. 
<<

ceki:
You write "when discardingThreshold != 0, but 51". Is 51 a typo or do  you actually mean the value 51?

ceki:
Anyway, I agree that when neverBlock is set to true, the value of discardingThreshold seems to be redundant with neverBlock=true. However, you may have a situation where the rate of arrival of non-discardable events (e.g. of level ERROR) is higher than the rate of service such that the buffer is full. In such a case, setting neverBlock to true can make sense.

Pro100AlexHell:
51 is 20% from 256, 51 is default discardingThreshold

PR #293, opened 2015-12-31
LOGBACK-1118: logback.groovy/Gaffer variable scope is broken

forsel:
Please have a look at http://jira.qos.ch/browse/LOGBACK-1118 for more details. Text without examples:

The normal variable scope as expected in Groovy scripts does not work properly in logback.groovy. Declared variables are accessible in all parts of the script, whereas fields (using @Field annotation) and undeclared variables are only accessible in the top-level code and not in any pre-defined closures.

I found that the cause of this is in the ComponentDelegate class. All logback configuration components are represented as closures and the ComponentDelegate class is set as delegate of each component. The delegation strategy is changed to DELEGATE_FIRST, so that component specific fields and methods can be handled by ComponentDelegate via the missingMethod and missingProperty methods. However, when a property or method is found to be component specific , the code logs an error and simply returns. Choosing strategy DELEGATE_FIRST, would mean that if the delegate cannot find the method/field, the owner should be checked next. This is not done, because of the return. Instead, ComponentDelegate should throw a MissingMethodException/MissingPropertyException. This signals to the Closure (class) code that it should continue looking for the missing method/field. And this is where the field or undeclared variable are then found.

Note that the strange behaviour for the hostname variable (a binding variable) as described in the documentation and also in LOGBACK-488 is caused by this as well. After applying the above fix, hostname simply resolves even when it is used in a closure.

I have made a fix where both return statements are replaced with throwing mentioned exceptions. As the 2 error logs are no longer necessarily applicable, I have removed them (gave too much clutter in my opinion). Instead, for the MissingPropertyException, the same string as for the error log is used and eventually, if the property is never found, the same message comes out as an exception. The MissingMethodException has a fixed constructor signature, therefore this is followed. If a method is used, that is not found as part of the component (e.g. appender, encoder, etc) and not present in the script, then the not-found method comes out as an exception.

In the same fix, I have updated 2 unit test input files as they referred to an unknown variable p. Before this would print out null, now it gives an exception back. Instead, I opted to print out the hostname variable instead.

Also, I have updated the site documentation by removing the extra HOSTNAME workaround and by adding a section that explains about variable scope.


forsel:
Rebased this pull request to release 1.2.3, which is the last release to support Groovy configuration for now. Note that 9 commits between release 1.2.2 and 1.2.3 are not on the master branch and therefore they show up in this pull request as well. Possibly, this pull request should be merged into branch brfanch_1.2.x instead?

PR #298, opened 2016-01-11
LOGBACK-1002: include() for Groovy configurations (alternative implementation)

forsel:
I have created an alternative implementation for the logback.groovy/Gaffer include() support that was mentioned LOGBACK-1002 and also implemented in pull request 257. I need include() support for my project and tried out the original proposed implementation. Although the original implementation seems very short and probably elegant in a 'Groovy' way (I'm new to Groovy, but a closure seems neat), I found that it had some shortcomings. Biggest issue was that included files were not scanned for changes. Also, the proposed implementation does not really try to follow the approach that was taken in the logback.xml (Joran) implementation. In my opinion, both Joran and Gaffer configuration should look very similar and it should be attempted to support the same featureset (except for the obvious added Groovy power), so that it is easy to move from one configuration to the other configuration.

I made the following update:
  - include method is added to ConfigurationDelegate, like all other DSL methods
  - include method supports File, URL (including resource URL) and String, where the latter is attempted to be resolved as URL, resource or File (in that order)
  - include method support optional flag to supress warning in case file to include cannot be found
  - logging is similar to that of Joran xml configuration, e.g. check for existing file/URL is done and warning is logged (if optional=false)
  - script parsing is done by GroovyConfigurator as before, so that 1 GroovyShell and 1 Binding is used for all scripts
  - by using 1 Binding, undeclared variables in the binding can be shared between scripts
  - in order to have script classes named after the script name, script parsing is done based on a URL instead of the direct script text
  - to have a simple, consistent implementation, logback.groovy itself is also parsed as a URL, the resulting class is no longer named Script1, but 'logback' after the filename
  - above 2 steps needed some changes in GroovyConfigurator, where the run(dslText) method was removed in favour of the run(url) method (direct script text seemed to be only used in 2 unit tests) 
  - the calling script class is accessible to the include script via a getParent method (parent field in ConfigurationDelegate)
  - the called/included script class is accessible to the calling script as a return value from the include method
  - 12 unit tests have been added, as much as possible similar to the extensive tests that were already present for Joran
  - extensive description for include method has been added to Chapter 12 of the manual, including a few extra example files under logback-examples
  - small unrelated update: groovy.transform.Field has been added to the ImportCustomizer, so that it is possible to create script class fields without needing to import this annotation (see also LOGBACK-1118)
  - small unrelated documentation update: add groovy.transform.Field to the list in the documentation

  
Please consider accepting and merging this proposed implementation (and preferably my other pull request regarding LOGBACK-1118 as well) as it makes Groovy-based logback configurations much more usable. Of course, any questions are welcome.


r04r:
Would love to see this functionality

mpsmith:
Me too... groovy support looks great except for this lack of include.

forsel:
Rebased this pull request to release 1.2.3, which is the last release to support Groovy configuration for now. Note that 9 commits between release 1.2.2 and 1.2.3 are not on the master branch and therefore they show up in this pull request as well. Possibly, this pull request should be merged into branch brfanch_1.2.x instead?

PR #302, opened 2016-02-23
Fix NPE on JMX reloadDefaultConfiguration when no configuration file is used

vpavic:
NPE occurs in use cases where Logback configuration is done programmatically. For example, Spring Boot uses Logback as default logging implementation and provides various means of customizing its default configuration. Enabling JMX configurator in such setup causes reloadDefaultConfiguration JMX operation to throw NPE.

This PR adds check to conditionally run configurator if the configuration file URL is indeed available.

PR includes a test case which reproduces the issue with current implementation. If needed, I can also provide a simple Spring Boot project to reproduce the issue.

Do I need to sign the CLA for this PR?

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/jmx/JMXConfigurator.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/jmx/JMXConfigurator.java
@@ -1,6 +1,6 @@
 /**
  * Logback: the reliable, generic, fast and flexible logging framework.
- * Copyright (C) 1999-2015, QOS.ch. All rights reserved.
+ * Copyright (C) 1999-2016, QOS.ch. All rights reserved.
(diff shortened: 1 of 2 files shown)


vpavic:
Closing & re-opening to force a new CI build.
The failed test passes for me locally.

ceki:
The CLA is not mandatory for this contribution as it is only few lines long. However, you are welcome to send me a singled CLA if you wish to do so.

vpavic:
@ceki I've sent you signed copy of the CLA according to instructions found in contribution guideline doc.

Regarding the PR, Travis CI failure seems unrelated to this change.

ceki:
@vpavic thank you for the cla. Travis CI failure is unlikely to be related to this change.

PR #305, opened 2016-03-04
SSL hostname verification

harbulot:
When [using SSL](http://logback.qos.ch/manual/usingSSL.html), there currently is no hostname verification: distinct servers with certificates issued by trusted CAs could impersonate each other.

This small patch implements hostname verification (using Java 7's "HTTPS" endpoint identification algorithm rules), and turns it on by default. It also adds a configuration to turn it off (`dontVerifyHostname`):

```
    <ssl>
        <trustStore>
            <location>classpath:/democa.jks</location>
            <password>demodemo</password>
        </trustStore>
        <parameters>
            <dontVerifyHostname>true</dontVerifyHostname>
        </parameters>
    </ssl>
```

(I realise this may break some existing configurations, but security by default is generally the right choice.)

This uses Java's [`SSLParameters.setEndpointIdentificationAlgorithm(String)`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html#setEndpointIdentificationAlgorithm-java.lang.String-), which is only available from Java 7. I have therefore implemented it using reflection, which rather unsatisfactory, but works. (I have also silenced the potential reflection exceptions, which is not great, but I wasn't sure how to handle this w.r.t. the rest of the code base.)
Hostname verification will still be missing when using Java 6.

Please note that Java's hostname matching rules follow RFC 2818 quite strictly. In particular, when using an IP address, that IP address MUST be in a Subject Alternative Name (of type IP address), not just in the CN. The documentation should probably reflect this. When using an IP address, instead of using:

```
keytool -genkey -alias server -dname "CN=my-logging-server" \
    -keyalg RSA -validity 365 -keystore server.keystore
```

the `-ext san=ip:<THE.IP.ADDRESS>` option should be used. For example (assuming the server address is `10.0.0.1`):

```
keytool -genkey -alias server -dname "CN=10.0.0.1" \
    -ext san=ip:10.0.0.1 \
    -keyalg RSA -validity 365 -keystore server.keystore
```

This requires `keytool` from Java 7 or above, but it can produce keystores usable with Java 6.

Feedback welcome, I don't mind modifying this patch if the pull request cannot be merged as it is.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurableSocket.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurableSocket.java
@@ -13,6 +13,7 @@
  */
 package ch.qos.logback.core.net.ssl;
 
+import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
 
 /**
@@ -60,4 +61,12 @@ public class SSLConfigurableSocket implements SSLConfigurable {
         delegate.setWantClientAuth(state);
     }
 
+    public void setDisableHostnameVerification(boolean disableHostnameVerification) {
+        if (disableHostnameVerification) {
+            return;
+        }
+        SSLParameters sslParameters = delegate.getSSLParameters();
+        sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
+        delegate.setSSLParameters(sslParameters);
+    }
 }
--- a/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurable.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurable.java
@@ -83,4 +83,5 @@ public interface SSLConfigurable {
      */
     void setWantClientAuth(boolean state);
 
+    void setDisableHostnameVerification(boolean disableHostnameVerification);
 }
--- a/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurableServerSocket.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurableServerSocket.java
@@ -60,4 +60,7 @@ public class SSLConfigurableServerSocket implements SSLConfigurable {
         delegate.setWantClientAuth(state);
     }
 
+    public void setDisableHostnameVerification(boolean disableHostnameVerification) {
+        // This is not relevant for a server socket.
+    }
 }
--- a/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLParametersConfiguration.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLParametersConfiguration.java
@@ -40,6 +40,7 @@ public class SSLParametersConfiguration extends ContextAwareBase {
     private Boolean wantClientAuth;
     private String[] enabledProtocols;
     private String[] enabledCipherSuites;
+    private Boolean disableHostnameVerification;
 
     /**
      * Configures SSL parameters on an {@link SSLConfigurable}.
@@ -54,6 +55,7 @@ public class SSLParametersConfiguration extends ContextAwareBase {
         if (isWantClientAuth() != null) {
             socket.setWantClientAuth(isWantClientAuth());
         }
+        socket.setDisableHostnameVerification(disableHostnameVerification != null ? disableHostnameVerification : false);
     }
 
     /**
@@ -239,4 +241,11 @@ public class SSLParametersConfiguration extends ContextAwareBase {
         this.wantClientAuth = wantClientAuth;
     }
 
+    public void setDisableHostnameVerification(Boolean disableHostnameVerification) {
+        this.disableHostnameVerification = disableHostnameVerification;
+    }
+
+    public Boolean isDisableHostnameVerification() {
+        return disableHostnameVerification;
+    }
 }
--- a/logback-core/src/test/java/ch/qos/logback/core/net/ssl/mock/MockSSLConfigurable.java
+++ b/logback-core/src/test/java/ch/qos/logback/core/net/ssl/mock/MockSSLConfigurable.java
@@ -27,6 +27,7 @@ public class MockSSLConfigurable implements SSLConfigurable {
     private String[] enabledCipherSuites = EMPTY;
     private boolean needClientAuth;
     private boolean wantClientAuth;
+    private Boolean disableHostnameVerification;
 
     public String[] getDefaultProtocols() {
         return defaultProtocols;
@@ -92,4 +93,11 @@ public class MockSSLConfigurable implements SSLConfigurable {
         this.wantClientAuth = wantClientAuth;
     }
 
+    public boolean isDisableHostnameVerification() {
+        return disableHostnameVerification;
+    }
+
+    public void setDisableHostnameVerification(boolean disableHostnameVerification) {
+        this.disableHostnameVerification = disableHostnameVerification;
+    }
 }


logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurableSocket.java
```diff
@@ -60,4 +64,24 @@ public void setWantClientAuth(boolean state) {
...
+             * to check whether it is available.
+             */
+            Method setEndpointIdentificationAlgorithmMethod = SSLParameters.class.getMethod("setEndpointIdentificationAlgorithm", String.class);
+            setEndpointIdentificationAlgorithmMethod.invoke(sslParameters, "HTTPS");
+            delegate.setSSLParameters(sslParameters);
+        } catch (NoSuchMethodException e) {
+        } catch (IllegalAccessException e) {
+        } catch (IllegalArgumentException e) {
+        } catch (InvocationTargetException e) {
+        }
```
ceharris: empty catch blocks allow disaster to propagate in a perhaps very unsafe way... better to wrap and rethrow as a `RuntimeException`

ceharris:
It's probably worth pointing out that this issue arises from the use of certificates issued by a non-local CA. In many cases, local CAs or self-signed certificates are being used, in which case hostname verification adds little benefit.

Nonetheless, I think it's a worthwhile feature.

harbulot:
> empty catch blocks allow disaster to propagate in a perhaps very unsafe way... better to wrap and rethrow as a RuntimeException. The only one that should occur legitimately is NoSuchMethodException, which should be silently ignored.

Yes, I agree, I'll try to update this PR over the next couple of days.

> It's probably worth pointing out that this issue arises from the use of certificates issued by a non-local CA. In many cases, local CAs or self-signed certificates are being used, in which case hostname verification adds little benefit.

That's true, the problem can be mitigated by using a specific trust store restricted to those local CAs. It mostly becomes a bigger problem when using larger trust stores (such as the default trust store), which can contain dozens of CA certs outside the organisation's control.

richardfearn:
I think it would be better if the parameter name wasn't negative, i.e. was called `verifyHostname`.

harbulot:
@richardfearn Is it just because `dont` doesn't sound good? In this case, `ignoreHostnameVerification` might sound a bit better, with the same behaviour.

`verifyHostname` is also feasible (even set to true by default when nothing is specified). My main reservation would be about handling incorrect configurations.

Indeed, the following would in fact treat `verifyHostname` as false:

    <parameters>
        <verifyHostname>yes</verifyHostname>
    </parameters>

This has to do with the way [`Boolean`](https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#Boolean-java.lang.String-) is parsed: 

>  The boolean returned represents the value true if the string argument is not null and is equal, ignoring case, to the string "true".
> Example: Boolean.parseBoolean("True") returns true.
> Example: Boolean.parseBoolean("yes") returns false.

It would be an easy mistake to make, and the user would assume it's configured securely, when it's not.

richardfearn:
Hmm, `<verifyHostname>yes</verifyHostname>` meaning that hostnames are *not* verified would be unfortunate.

I don't like parameters with negative names - I think `verifyHostnames` (true or false) is better than `dontVerifyHostname` (true means no don't verify, false means yes). Easier to understand if it's just `verifyHostnames` - is it doing that, or not.

harbulot:
@richardfearn What do you think of `ignoreHostnameVerification` (it seems to be used by a few other tools), or `disableHostnameVerification`, or even perhaps `trustAnyHostname` (although that might be more ambiguous in what it does)? It seems to sound less negative, and possibly clearer.

Whatever the name, I think having something that needs to be `true` to disable hostname checking is probably a little bit better (a) to avoid that `"yes"==false` problem, and (b) to make it a more "concious" choice from the user (disabling a security feature on purpose).

I'd probably lean towards `ignoreHostnameVerification`. (That said, I'm open to the reverse behaviour, with `verifyHostname` if needed, despite its drawbacks, depending on other feedback too.)

richardfearn:
`disableHostnameVerification` seems better than `ignoreHostnameVerification`.

ceharris:
`disableHostnameVerification` seems to be best compromise among the various forces at work here

harbulot:
I've just updated the PR. The settings are now:

    <parameters>
        <disableHostnameVerification>true</disableHostnameVerification>
    </parameters>

The value has to be "true" (case insensitive) for hostname verification to be disabled. Otherwise, it's enabled.

logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurableSocket.java
```diff
@@ -60,4 +64,28 @@ public void setWantClientAuth(boolean state) {
...
+             */
+            Method setEndpointIdentificationAlgorithmMethod = SSLParameters.class.getMethod("setEndpointIdentificationAlgorithm", String.class);
+            setEndpointIdentificationAlgorithmMethod.invoke(sslParameters, "HTTPS");
+            delegate.setSSLParameters(sslParameters);
+        } catch (NoSuchMethodException e) {
+            // We don't do anything when the method is not available.
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        } catch (IllegalArgumentException e) {
+            throw new RuntimeException(e);
```
tony19: Codacy Issue found: [Avoid throwing raw exception types.](https://www.codacy.com/app/tony19/logback/file/2371120379/issues/source?bid=3222618&fileBranchId=3222658#l86)

tony19: Codacy Issue found: [Avoid throwing raw exception types.](https://www.codacy.com/app/tony19/logback/file/2371120379/issues/source?bid=3222618&fileBranchId=3222658#l86)

harbulot: @tony19 What else do you suggest in this particular case?

tony19: @harbulot I recommend [`LogbackException`](http://logback.qos.ch/apidocs/ch/qos/logback/core/LogbackException.html).

(btw, I didn't mean to jump in on this review. I was merely experimenting with Codacy and did not expect it to comment on all PRs in all projects. oops)

ceki: :-)

With the exception of FindBugs in "high priority" mode, I am yet to encounter an automatic tool with a reasonable signal to noise ratio. Codacy does not seem to be an exception to the rule.

tony19: I don't think you've looked close enough :smile: Like FindBugs, Codacy allows configuration of checks:

screen shot 2016-04-07 at 12 02 03 pm

I'm not sure if you were alluding to this issue here as noise, but I think Codacy made a good point.

ceki: I have not looked at this particular issue but was responding to your comment.

andreikom:
Could we kindly revive this? What needs to be done so that this can get merged?

ceki:
@andreikom  Could you create a jira issue please?

andreikom:
@ceki Thx for the prompt response - https://jira.qos.ch/browse/LOGBACK-1574

Btw - is there a way to disable the Socket appenders all together? I tried looking up both in the documentation (http://logback.qos.ch/manual/appenders.html#SocketAppender) and the code however I could not pinpoint such an option.

ceki:
@andreikom You mean for security reasons?

andreikom:
Yes

harbulot:
I've just had a quick look at this PR again (sorry, can't even remember submitting it 5 years ago :-) )

It seems when I wrote this patch, I took into account the fact it could run on Java 6, so I was using reflection to try the Java 7+ features.

As it seems the current version only supports Java 8+ anyway, this could probably be simplified to something like this (not tried), which would also address the comments about Exceptions above:

```java
    public void setDontVerifyHostname(boolean dontVerifyHostname) {
        if (dontVerifyHostname) {
            return;
        }
        SSLParameters sslParameters = delegate.getSSLParameters();
        sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
    }
```

arkangelboss-github:
Thank you @andreikom, @harbulot and @ceki for picking this change up again. I started checking security issues a few weeks ago and, after inquiring with JFrog, it seems that they are reporting a Security concern on Logback because of this.

When I inquired about the report below, they pointed me explicitly to this PR, and it would be great to see it addressed.

The connected Security report:
```
Summary: logback SSL Certificate Validation Failure MitM Spoofing 
Severity: MEDIUM
Description: logback contains a flaw as X.509 certificates are not properly validated. By spoofing the  TLS/SSL server via a certificate that appears valid, an attacker with the  ability  to  intercept  network traffic (e.g. MitM, DNS cache poisoning) can disclose and optionally manipulate transmitted data.
Type: SECURITY
Provider: JFrog
Issues: 4.0/CVSS:2.0/AV:N/AC:H/Au:N/C:P/I:P/A:N
Edited: 2021-04-15T09:22:04Z
Created: 2019-05-02T00:00:00.297Z
```

ivanopagano:
Sorry to bother you, any progress on this issue?
Could it be part of the next release or you see it as improbable?

logback-core/src/main/java/ch/qos/logback/core/net/ssl/SSLConfigurableSocket.java
```diff
@@ -60,4 +64,28 @@ public void setWantClientAuth(boolean state) {
...
+            /*
+             * SSLParameters.setEndpointIdentificationAlgorithm(...)
+             * is only available from Java 7, so we use reflection
+             * to check whether it is available.
+             */
+            Method setEndpointIdentificationAlgorithmMethod = SSLParameters.class.getMethod("setEndpointIdentificationAlgorithm", String.class);
+            setEndpointIdentificationAlgorithmMethod.invoke(sslParameters, "HTTPS");
+            delegate.setSSLParameters(sslParameters);
+        } catch (NoSuchMethodException e) {
+            // We don't do anything when the method is not available.
```
elruwen, line 82: Is that really a good idea? I don't set disableHostnameVerification. But if I don't have that method, the hostname verification silently fails. I personally would prefer another RuntimeException. Or at least a warning printed. In the current situation it is impossible for me as a user to know if hostname verification is actually enabled or if it silently failed.
If it throws an exception because it is not supported, then the hostname verification can be still explicitly disabled.

Alos:
It would be great if this was submitted so that the Spring folks can pick up this update.

ceki:
This PR will be committed shortly using a parameter called "hostnameVerification". 

Having said that, apart from people complaining about the vulnerability discussed, I wonder if people care about SSLAppender/SSLServer as in "we actually use it".

hazendaz:
Hi Ceki, to answer your question as it's a bit rhetorical.  The tools involved don't really care if it's used or not.  It's vulnerable and unfortunately it puts everyone restricted to use these tools in a bad position especially when the initial PR existed for 5 years.  A fix with flag even if not properly used will shut the tools up even if issue is still present.  It's silly for sure but desperately needed.  Because of this issue we have been forced to use 1.2.6 instead of some 1 3.0 due to missing code there as you have documented. Stuck on 1 3.0 alpha 4.  And there has been talk of replacing with log4j2 (yuck) or forking and applying the fix (yuck).  So any fix at this point will let us remain at least on 1.2.x.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

ceki:
Fixed in 626c773 (1.2 branch)

1.2.7 released earlier.

harbulot:
@ceki The patch I'd written was tweaked a little bit, and I don't think it fixes the problem.

A key line in the initial patch was:

```java
socket.setDisableHostnameVerification(disableHostnameVerification != null ? disableHostnameVerification : false);
```

This was replaced with

```java
if(hostnameVerification != null) {
    addInfo("hostnameVerification="+hostnameVerification);
    socket.setHostnameVerification(hostnameVerification);
}
```

Those are unfortunately not equivalent. For hostname verification to be enabled by default, `socket.setHostnameVerification(` **must** always be called, with `false` if needs be.

Sascha82:
Gents,

did I get it right, that this issue is not fixed with Release 1.2.7? Is there a chance to do another release in the next days? If not, is there any chance, that we can fix it with our own developers?

Thanks and best regards,
Sascha

hazendaz:
The fix as it is requires that user set the host name verification to true in their configuration on their own.  Its not set by default so its vulnerable unless setup but then only matters if you are using that feature anyways.  Not sure if this was accepted as fixed or not.  Synk still was showing it vulnerable over the weekend but that was because they marked not fixed all versions.  Personally regardless of requiring manual setup or not, the vulnerability IMHO was fixed.  But I would agree it probably should be the default and let those that don't care about security turn it off.

harbulot:
@hazendaz I'd generally err on the side of caution and suggest this vulnerability has not be fixed.

From my own general experience, developers/users generally assume that just using SSL/TLS will secure the communication, without looking further into which further settings need to be enabled. That's where the security vulnerability generally lies unfortunately.

Put it another way: if common browsers didn't perform hostname verification by default (even if there was an option to enable it), I'm fairly sure someone would flag this as a security vulnerability.

The state of affairs has improved a little since this 2012 paper: ["*The most dangerous code in the world: validating SSL certificates in non-browser software*" by M. Georgiev, S. Iyengar, S. Jana, R. Anubhai, D. Boneh, and V. Shmatikov](https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf), but probably not quite enough yet:

> Developers often misunderstand security guarantees provided by SSL libraries. For example, JSSE (Java Secure Socket Extension) has multiple interfaces for managing SSL connections. The “advanced” SSLSocketFactory API silently skips hostname verification if the algorithm field in the SSL client is NULL or an empty string rather than HTTPS. This is mentioned in passing in the JSSE reference guide, yet many Java implementations of SSL-based protocols use SSLSocketFactory without performing their own host-name verification.

Essentially, users generally have a fair expectation that the tool is going to do the right thing out of the box (once SSL is enabled): the vulnerability comes from the fact that it doesn't.

Sascha82:
@hazendaz & @harbulot: Thanks a lot for your feedback!

One option could be to not use SSL for logback. When I got it right from our Developers, we only use logback to store the logs from our asset. Looks like there is not much benefit to use SSL in this context. Do you have any use cases for the usage of SSL in this context? 

Thanks again for your valuable support.

Best regards,
Sascha

harbulot:
@Sascha82 This issue isn't really about how you store your logs, it's rather about how Logback sends them to a remote server that stores them. If you're only using a `FileAppender` locally, then none of this matters.

This is mainly for people sending their logs to a remote server using the `SSLSocketAppender` or something like [logstash-logback-encoder](https://github.com/logfellow/logstash-logback-encoder)'s `net.logstash.logback.appender.LogstashTcpSocketAppender`, which uses Logback's `SSLConfiguration`.

I think this issue has just gained attention after so long because it was recently given a vulnerability number in Snyk, but whether the upgrade is just about making people's Snyk dashboards happy or whether those people actually use the feature is an open question. That's probably what @ceki was questioning earlier:

> I wonder if people care about SSLAppender/SSLServer as in "we actually use it".

Sascha82:
@harbulot Thanks a lot for clarification!

PR #307, opened 2016-03-15
LOGBACK-1071: Java 8 compact3 profile compatibility.

urechm:
Copied necessary classes from java.beans/sun/com.sun packages directly into logback-core code in order to be java 8 compact3 compatible.


ceki:
Given that the license of packages under java.beans is GNU whereas logback is licensed under EPL and/or LGPL 2.1, copying these files would imply a change in license. I am sorry but this PR cannot be accepted.

maxurech:
Hmm, I am not a lawyer, but we do not actually change the source code, we just repack it...

ceki:
@maxurech My comment above was not very clear. I just modified it.  Copying said files would imply a change in license _for logback itself_.

PR #309, opened 2016-03-28
check if of CompositeConverters handle exceptions

epishkin:
This fix makes `EnsureExceptionHandling` check that if there is a `CompositeConverter` in a chain it can handle exceptions. Without this fix you have to add `%nopex` to the logging pattern if you do something like `%replace(ex{short}){'\w', ''}`.

I don't have tests for this fix yet. Could you point me where to find tests for `EnsureExceptionHandling`?
I will be glad to write additional tests for this.

thanks in advance,
Oleksii


ceki:
Can you please create a Jira for this as well?

PR #315, opened 2016-04-07
Add a StatusListener that logs to stderr/stdout based on status

raboof:


ceki:
I guess it makes sense under certain circumstances. However, I think it is up to the user to write a custom status listener if the circumstances mandate it.

raboof:
AFAICS this is currently impossible to do (elegantly), because `getPrintStream` doesn't have the `status` parameter.

Even if `getPrintStream` would have that parameter, putting the `OnStdStreamsStatusListener` in my project would mean adding logback as a compile-time dependency rather than a run-time one - which is probably usually not a big deal, but another hurdle to take.

So for these reasons I'd love to see `OnStdStreamsStatusListener` merged into logback itself. If you feel this is unacceptable, fair enough, feel free to close this PR :).

ceki:
Interesting point about depending on logback.

PR #316, opened 2016-04-08
Support unset MDCFilter value

msteiger:
Currently, the `MDCFilter` fails with an `NullPointerException` if the filter value is not set. I think that filtering unset MDC values is a reasonable use case and fixed the issue by replacing the equals check in analogy to `Objects.equals` from Java 7.

The unit tests verify the correctness of the new behavior.


ceki:
Can you please create a jira issue for this PR?

msteiger:
Done! See http://jira.qos.ch/browse/LOGBACK-1165

PR #317, opened 2016-04-09
LOGBACK-1049 fix incorrect variable name

cmayne-r7:
See http://jira.qos.ch/browse/LOGBACK-1049


cmayne-r7:
Closing due to inactivity

PR #321, opened 2016-05-09
fix for LOGBACK-1166: totalSizeCap not work as expected

asabbi:
Fixes: http://jira.qos.ch/browse/LOGBACK-1166
TimeBasedArchiveRemover.capTotalSize() was preserving 2 days of logs instead of 2 log files. This seems to be the intended behavior as the name UNTOUCHABLE_ARCHIVE_FILE_COUNT suggests. Made changes so that only 2 files are preserved.


ceki:
This PR is probably better than my latest solution.

asabbi:
@ceki Should I revert your changes to TimeBasedArchiveRemover and the unit tests so that this PR can be merged?

ceki:
No, the change can be applied manually.

PR #325, opened 2016-06-02
Allow reporting of logging loss in AsyncAppenderBase

erikvanoosten:
Made `AsyncAppenderBase.isQueueBelowDiscardingThreshold` public to facilitate reporting loss of log events in a health check.

For example, with this change one could now call `isQueueBelowDiscardingThreshold` from a dropwizard metrics health-check.


erikvanoosten:
The failing travis build has _no_ relation whatsoever with this PR.

erikvanoosten:
BTW, a more serious solution is to have a public counter with the number of dropped messages.

rtyley:
I encountered the same problem today, and created https://jira.qos.ch/browse/LOGBACK-1436 to track it.

PR #327, opened 2016-06-16
LOGBACK-1173 Remove key on stop() in FileAppender

kden:
Initially I didn't recognize that separate maps were being updated in FileAppender and RollingFileAppender.  744c6e1 corrects that.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/FileAppender.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/FileAppender.java
@@ -132,6 +132,14 @@ public class FileAppender<E> extends OutputStreamAppender<E> {
             super.start();
         }
     }
+    
+    @Override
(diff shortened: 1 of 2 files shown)


HomMarkHunt:
+1

mpderbec:
Seems like this has been addressed in latest logback code. Can this PR be dropped?

```
    @Override
    public void stop() {
        super.stop();
        
        if (rollingPolicy != null)
            rollingPolicy.stop();
        if (triggeringPolicy != null)
            triggeringPolicy.stop();

        Map<String, FileNamePattern> map = ContextUtil.getFilenamePatternCollisionMap(context);
        if (map != null && getName() != null)
            map.remove(getName());

    }
```

kden:
Addressed as above.  Closing.

PR #331, opened 2016-07-17
LOGBACK-783 - add bzip2 support

mgoellnitz:
Since the operations guys around me prefer to have bzip2 in use for archived / compressed logs, I thought my logging framework should support this. So I wanted to support the [discussion](http://jira.qos.ch/browse/LOGBACK-783) with a small pull request:
Logfile rotation patterns ending with bz2 will result in rotated log files compressed with bzip2.

### DIFF
--- a/logback-core/pom.xml
+++ b/logback-core/pom.xml
@@ -40,6 +40,12 @@
       <optional>true</optional>
     </dependency>
     <dependency>
+       <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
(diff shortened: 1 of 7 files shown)


diroussel:
This is failing CI, but that failure is affecting other PRs, so probably not related?
Should the new maven dependency be made optional?
Should there be updates to the documentation?

mgoellnitz:
The (now) failing CI should not be related... The dependency should better be optional, true. And: Documentation related to the compression feature needs an extension mentioning bzip2.

PR #332, opened 2016-08-10
Allow OutputStream and Encoder to be set on the Appender in any order

guenhter:
Currently it is necessary to set the encoder before the outputStream is set on the appender, to ensure that the `OutputStreamAppender#encoderInit` sets the outputStream on the encoder. 

This is now eliminated by calling the `encodeInit` Method also when the encoder is set.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/OutputStreamAppender.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/OutputStreamAppender.java
@@ -227,6 +227,8 @@ public class OutputStreamAppender<E> extends UnsynchronizedAppenderBase<E> {
 
     public void setEncoder(Encoder<E> encoder) {
         this.encoder = encoder;
+
+        encoderInit();
     }
 
 }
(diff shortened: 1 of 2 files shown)


diroussel:
@guenhter any thoughts on the CI failure?

guenhter:
@diroussel I tried the commit before my changes -> the same test fails
I just rebased my fork to the latest master and the build is fine. Maybe you can just retrigger the build?

guenhter:
@diroussel Seems to work now.

ceki:
I can't remember changing anything on this regard.

guenhter:
But it seems to work now. Can this be merged?

ceki:
Encoder code changed radically in 1.2.0. Does it still make sense to merge this PR?

guenhter:
IMO it still would make sense, because the `encoderInit` method still has dependencies to the encoder and the outputStream.
BUT after another glimpse at the `setOutputStream`-Method I saw the following code:
```
if (encoder == null) {
    addWarn("Encoder has not been set. Cannot invoke its init method.");
    return;
}
```

I think this can be understood that way, that the encoder must be set before the outputStream is set.
So I will close this one.

PR #336, opened 2016-10-06
LOGBACK-1189: detect previous copyAttributeMap calls

markelliot:


markelliot:
I ran into http://jira.qos.ch/browse/LOGBACK-1189 in a personal project, but noticed that at least a few others have also run into this problem quite recently:
https://github.com/logstash/logstash-logback-encoder/issues/94
https://github.com/dropwizard/dropwizard/issues/1699

evnm:
As illustrated by the Dropwizard issue that @markelliot linked to, we've been debugging a consequence of this mutable state issue for months now. For more information, please see [my explanation sent to logback-user@](http://mailman.qos.ch/pipermail/logback-user/2016-September/004800.html). Would really appreciate some guidance from the Logback maintainers.

markelliot:
@qos-ch/logback: any chance you can take a look at this? would love to get this simple change merged and released.

nickbabcock:
I'm not familiar with the codebase, so I could be totally off, but is there still a race condition where two method calls (in parallel) could see that the map is null and both attempt to fill it?

markelliot:
Happy to mark this method as synchronized or synchronize more explicitly on attributeMap, but as far as I can tell the other methods in the class would also be susceptible to multiple threads accessing this object. 

The fix here seeks primarily to make the method idempotent even for single threads, like all of the other methods.

nickbabcock:
You're right, looks like no other synchronization primitives are used in the class. Carry on 😋

markelliot:
I've rebased this to latest master, what else can I do to help get this merged?

markelliot:
@ceki what work is required to get this change or an equivalent fix to the underlying issue merged and released?

ceki:
@markelliot Thank you for following up on this issue.

PR #337, opened 2016-10-26
Adds a getSingleton method expected by SLF4J

bric3:
SLF4J introduced the change in 1.7.14, see https://github.com/qos-ch/slf4j/blob/v_1.7.21/slf4j-api/src/main/java/org/slf4j/MDC.java#L88-L104
The `StaticLoggerBinder` is up to date but not `StaticMDCBinder`


bric3:
It seems the CI failure is [time related](https://travis-ci.org/qos-ch/logback/builds/170809890#L3654)

diroussel:
Can the build be retriggered?

PR #339, opened 2016-11-04
Add MDC support to logback-access (LOGBACK-1016)

pgr0ss:
- Add MDC support to AccessEvent (mirroring the code in LoggingEvent)
- Move LogbackMDCAdapter and MDCConverter to core so it can be shared by both
- Refactor MDCConverterTest to have fewer dependencies

This is the first time I've gone through the logback code, so please let me know if I didn't do something correctly or there is a better way to do something. Thanks!

### DIFF
--- a/logback-access/pom.xml
+++ b/logback-access/pom.xml
@@ -28,6 +28,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.tomcat</groupId>
       <artifactId>tomcat-catalina</artifactId>
       <scope>compile</scope>
--- a/logback-access/src/main/java/ch/qos/logback/access/PatternLayout.java
+++ b/logback-access/src/main/java/ch/qos/logback/access/PatternLayout.java
@@ -44,6 +44,7 @@ import ch.qos.logback.access.pattern.SessionIDConverter;
 import ch.qos.logback.access.pattern.StatusCodeConverter;
 import ch.qos.logback.access.pattern.ThreadNameConverter;
 import ch.qos.logback.access.spi.IAccessEvent;
+import ch.qos.logback.core.pattern.MDCConverter;
 import ch.qos.logback.core.pattern.PatternLayoutBase;
 import ch.qos.logback.core.pattern.color.*;
 import ch.qos.logback.core.pattern.parser.Parser;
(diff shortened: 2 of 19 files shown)


dimas:
Came across your PR by accident, thought worth telling you... I am not sure your change is a very good idea for two reasons:
1. bringing slf4j into logback-core seems wrong. It should not be already tied to any of the logging facades at that point
2. the whole idea of using MDC with logback-access, I am not sure it is going to work. From what I saw, things like MdcInsertingServletFilter do MDC.clear() in the "finally" block. And logback-access gets the request for logging AFTER the entire filter chain has completed. Which means by the moment request arrives for logging, it will already lose all MDC data. (Which can be different on different containers of course, my experience was with Spring & Jetty). I believe the only proper way of logging something with the request is by setting request attribute.

Cheers

ceki:
I agree with @dimas. SLF4J cannot/should not become a dependency of logback-access.
The second point mentioned by @dimas is even more cogent.

pgr0ss:
@dimas Thanks for the feedback. FWIW, this change does work with dropwizard 1.0.2, so the MDC is not getting cleared out. But I can't speak to other containers. I'm not familiar with request attributes, but I can look into it and give it a try.

pgr0ss:
@dimas, @ceki I tried to switch my code to use request attributes instead of the MDC, but I couldn't get it working. For example, I made a request filter that looked roughly like this:

```java
class RequestIdFilter implements ContainerRequestFilter {

  @Context
  private HttpServletRequest servletRequest;

  @Override
  public void filter(ContainerRequestContext containerRequestContext) throws IOException {
    servletRequest.setAttribute("requestId", "HELLO");
  }
}
```

And then I used that attribute in a pattern:

```java
private static class MyRequestLayout extends LogbackAccessRequestLayout {
    public MyRequestLayout(LoggerContext context, TimeZone timeZone) {
        super(context, timeZone);
        setPattern("INFO  [%date{ISO8601}] [%requestAttribute{requestId}] " + getPattern());
    }
}
```

But it prints as `[-]`. I stepped through with the debugger, and I see the attribute being copied in `AccessEvent.copyAttributeMap`. However, when `getAttribute` is called (in another thread I think), the map is empty. Are these getting cleared out somewhere? Is there something obvious I'm doing incorrectly?

Thanks for any help you can give.

pgr0ss:
Dropwizard recently brought back request logging with logback-classic (https://github.com/dropwizard/dropwizard/pull/1813), so this issue isn't as pressing for me. I'm still not sure why my request attribute example didn't work, but I can close this PR since it's not the direction you want to go.

markelliot:
The request attribute stuff fails for you because Dropwizard is susceptible to the bug fixed by #336.

markelliot:
(details of why on the linked JIRA ticket)

PR #341, opened 2016-11-08
Fix NPE

wuwen5:


huxi:
Patch was simple enough that I felt competent to just merge it.

PR #344, opened 2016-11-23
LOGBACK-1225 Make setContextMap() be happier about null

dimas:


logback-classic/src/main/java/ch/qos/logback/classic/util/LogbackMDCAdapter.java
```diff
@@ -13,13 +13,13 @@
  */
 package ch.qos.logback.classic.util;
 
+import org.slf4j.spi.MDCAdapter;
```
diroussel: I don't think you should be changing the order of imports

logback-classic/src/test/java/ch/qos/logback/classic/util/LogbackMDCAdapterTest.java
```diff
@@ -13,18 +13,14 @@
...
+import ch.qos.logback.core.testUtil.RandomUtil;
+import org.junit.Test;
 
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 
-import org.junit.Test;
-import ch.qos.logback.core.testUtil.RandomUtil;
+import static org.junit.Assert.*;
```
diroussel: Again, please honour existing code formatting style for imports

PR #346, opened 2016-12-05
fix tomcat byte send for chunked encoding

ckoutsouridis:


ceki:
Is there a Jira issue for this PR? If not, can you please create one?

ckoutsouridis:
Created ticket and updated the title as well

ceki:
Do you know when the "if (length <= 0)" branch would be activated?

ckoutsouridis:
response.getBytesWritten(false); returns 0 when no bytes are written, like in a regular 304 response status. I had to add that last check so that when no bytes are written, the method returns -1 in order for the converter (ContentLengthConverter.java) to properly set it to -. I noticed that if the method returns 0, the 0 is written in the logs instead of the dash.

ceki:
Unless we can explain the "request.getAttribute( Globals.SENDFILE_FILE_START/END_ATTR)" part, we should not keep it.

ckoutsouridis:
those attributes are used by tomcat's DefaultServlet when
1) serving static resources
2) send file is enabled (default is enabled)
3) file content is more that the configured one (default about 50kb i think)

The response headers in such a case will look like:
```
Accept-Ranges:bytes
Content-Length:10140000
Date:Tue, 06 Dec 2016 12:12:53 GMT
```
i believe we could do a workaround when getBytesWritten() returns <=0 to fallback to response.getContentLenght() which is populated as well in the above case.

Or to keep the change minimal we could first check response.getContentLenght()  and then fallback to: getBytesWritten()

PR #351, opened 2016-12-19
LOGBACK-1240 Log files aren't deleted on startup with hourly rotation

asaarilahti:
Remove casting to int in calculation of periodBarriersCrossed in RollingCalendar and add a test. Casting caused the the value to become negative resulting in no archives being deleted.


kylixs:
why this bug is not fix?

ceki:
In https://jira.qos.ch/browse/LOGBACK-1240 added to the list of issues to be fixed for the next release.

PR #353, opened 2016-12-27
fix LOGBACK-1231: totalSizeCap only work for about 2GB file size beca…

superyizhi:
fix LOGBACK-1231


gracew:
looks like this can be closed, it was fixed in https://github.com/qos-ch/logback/commit/2240d4253c9f3fa28497e74f34be3bb1edc73078 which is in 1.2.0.

PR #356, opened 2017-01-23
Fix LOGBACK-1247.

jakber:
AsyncAppenderBase no longer drops events when the thread is interrupted.
Tested by changing the existing test for PULSE-910 to also verify that the event was logged.

I'm getting some test failures in DefaultSocketConnectorTest, but those happen both with and without my changes.


jakber:
Since the tests passed on Travis I guess the DefaultSocketConnectorTest failures I saw locally were due to my environment.

ceki:
@jakber   Manually merged per https://github.com/qos-ch/logback/commit/7d62891efc00cbe40f8f13b0621628343e447b7c. Although I tried, the merge did not retain your name. Sorry about that.

PR #357, opened 2017-01-29
fix LOGBACK-1194: scanPeriod attribute required for auto-reload to work

fbuechler:
Use a default scanning period of one minute if scan="true" but scanPeriod is not set.

I added a test that checks if there are any scheduled tasks in the context which seems to work fine. But it feels like a rather indirect way to test if the default is used.


fbuechler:
I've tried to figure out why UniqueFileTest is failing but I can't reproduce the issue locally.

ceki:
@fbuechler This is a very good PR. Thank you for submitting. I have incorporated most of it manually as existing code (in master not this PR) has other warts.

ceki:
Reopened to allow answers.

PR #359, opened 2017-02-03
support MDC in access logs

doumdoum:
Hi,
I've added MDC support for access logs.

Regards,

/DV

### DIFF
--- a/logback-access/pom.xml
+++ b/logback-access/pom.xml
@@ -23,6 +23,11 @@
     </dependency>
     <dependency>
       <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
       <artifactId>logback-core</artifactId>
       <type>test-jar</type>
       <scope>test</scope>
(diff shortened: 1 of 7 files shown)


ceki:
thank you for your contribution.  I'm afraid having logback-access depend on slf4j-api is unacceptable. From a cursory look at your code, it is not clear why you need the slf4j-api dependency.

doumdoum:
Hi ceki,
Thx for you quick review.
The problem comes from the fact that the MDC api is provided in the slf4j-api package.
I wanted to mode the MDC related stuff to logback-core but the dependency to slff4j-api is still mandatory.
Any hint on how to resolve that ?

Thx,
/DV

doumdoum:
Hi Ceki,
I've made some code changes according to your review.

Sharpie:
This looks like it provides the requested abstraction from SLF4J via the MDCAdapter class.

@ceki Anything else that needs to be done on this PR? The ability to use MDC items in a logback-access pattern would be really helpful.

dfredell:
This would be a very nice feature. I'm coming from the dropwizard community and there is interest from them too https://jira.qos.ch/browse/LOGBACK-1016

It looks like @ceki hasn't been active on Git Hub since March 2018. Is there anyone else who can approve pull requests? Otherwise we might just try building and bundling @doumdoum's branch internally.

ceki:
So the idea is to use an MDCAdapter if available on the class path?

dskeppstedt:
Any news on this?

PR #361, opened 2017-02-23
[pom] Upgrade junit to 4.12 and exclude old version of hamcrest

hazendaz:


hazendaz:
Fix coming shortly...was recomfirming old issue and my local isn't building for some time with a bad test case.

hazendaz:
closing will reopen new one.  need to fix some things due to mockito movement to managed parent.

PR #139, opened 2013-08-02
Added the ability to override the hostname when using SyslogAppender

leonsodhi:
Added a hostname tag that will allow someone to easily override the hostname sent out in the syslog packet. This is particularly useful on services like EC2 where the hostname can be quite cryptic, i.e., made up of the MAC or IP address.

PR #364, opened 2017-03-05
allowing for layout to be overridden

vrolijken:
I'm trying to change the behaviour of the SyslogAppender. One of the things I'd like is short hostnames and I’m not the only one: 
http://logback.10977.n7.nabble.com/Syslog-Appender-Short-Host-Name-td1102.html
https://github.com/qos-ch/logback/pull/139/files
https://jira.qos.ch/browse/LOGBACK-1011 
This change allows me to override the layout without affecting current functionality, so current user should not be impacted.


vrolijken:
All checks have passed. Can I do anything further to get this accepted?

PR #365, opened 2017-03-12
test: add test on equals of FileNamePattern against null value

danglotb:
Hello,

I propose this changes to specify the [equals()](https://github.com/qos-ch/logback/blob/master/logback-core/src/main/java/ch/qos/logback/core/rolling/helper/FileNamePattern.java#L96) method of `ch.qos.logback.core.rolling.helper.FileNamePattern` against null value


danglotb:
Hello, have you reviewed the pull request? 
Thanks.
-- Benjamin.

ceki:
I am curious as to why FileNamePattern.equals(null) returning false is of any interest.

PR #366, opened 2017-03-13
Fixed typo within constant about incorrect attribute for parameters

andyfeller:
Constant referenced incorrect attribute name in the event that a parameter action was misconfigured as the constant for the attribute in question is actually **value**.


ceki:
Thank you for this patch

PR #368, opened 2017-03-19
Added the feature to update mime message before sending the email

tetv:
Useful to extend the SMTPAppender class in order to add some Email Headers such as X-Priority, In-Reply-To, ...


ceki:
Why is there a need to pass CyclicBuffer to updateMimeMsg()?

tetv:
Because it can be useful to calculate for example the highest level of all messages and do something like:
```groovy
Level highLevel = cb.asList().stream().map(ILoggingEvent::getLevel).max(comparingInt(Level::toInt)).orElse(Level.DEBUG);
mimeMsg.setHeader("X-Priority", convertToPriorityString(highLevel));
```

ceki:
I see. Please create a jira issue for this PR. Otherwise, it will fall through the cracks.

tetv:
Done: LOGBACK-1284

PR #371, opened 2017-04-06
LOGBACK-1260 New logic to get JDK version

reinhapa:
Fixes issue [LOGBACK-1260](https://jira.qos.ch/browse/LOGBACK-1260)


ceki:
@reinhapa Would it be possible to change the logic to use an initial test to see of the java.version begins with "1.", if so extract the second sequence of digits, and if it does not start with "1." and extract the fist sequence of digits?

reinhapa:
@ceki is there an performance issue with the current solution besides that I could make the actual pattern static final?

reinhapa:
@ceki updated the logic using a alternative parse logic without regular expression pattern matching, that seems to me simple enough supporting you're request too :-)

PR #374, opened 2017-05-22
LOGBACK-1302: Throwable is trimmed twice which causes bad message

gagoman:
Closes https://jira.qos.ch/browse/LOGBACK-1302

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java
@@ -290,6 +290,12 @@ public class LoggingEvent implements ILoggingEvent {
             return formattedMessage;
         }
         if (argumentArray != null) {
+            if (throwableProxy != null && argumentArray.length > 0) {
+                Object possibleThrowable = argumentArray[argumentArray.length - 1];
+                if (possibleThrowable instanceof Throwable) {
+                    argumentArray[argumentArray.length - 1] = possibleThrowable.toString();
+                }
+            }
             formattedMessage = MessageFormatter.arrayFormat(message, argumentArray).getMessage();
         } else {
             formattedMessage = message;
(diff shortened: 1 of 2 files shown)


fenik17:
I think problem in the MessageFormatter. This [commit](https://github.com/fenik17/slf4j/commit/c3751df9afec9b412de6f888a97397af67063717) introduced a change in behaviour: code `log.info("{}", throwable)` began printing stack trace instead of throwable.toString(). IMHO, a new behaviour is incorrect.

gagoman:
This is a different issue. I have 2 last args as `Throwable`. I'm fine with last one printed as stack-trace. I want one before to be printed as it is or as expected (via `Formatter`, any advice is appreciated here), but not to be ignored. Please take a look at https://github.com/qos-ch/logback/pull/374/commits/4c951b8f6edb4c4dbee5989af6591c2be520a38d#diff-d22929da05e45e6ec8b3f754bac048efR93

fenik17:
I'm agree with you, that it's a bug. Look [at this line](https://github.com/fenik17/slf4j/commit/c3751df9afec9b412de6f888a97397af67063717#diff-f99407dbe36d6d3e23534b6e5701b231R171): the last array element cuted if it is Throwable,  even if the message template contains braces `{}` for it. 
So, I believe that the problem is in slf4j's `MessageFormatter`.

gagoman:
Yes, I thought that `MessageFormatter` is working as designed. There is an PR for it already: https://github.com/qos-ch/slf4j/pull/166

Should we close this one and continue there? Or should I cleanup code and leave tests only to show an issue and wait https://github.com/qos-ch/slf4j/pull/166 to be merged?

fenik17:
> There is an PR for it already: qos-ch/slf4j#166

Yes that's it. I have not seen this PR before...

> Should we close this one and continue there?

Yep. Just close this issue (and jira issue) as duplicate.

PR #378, opened 2017-06-20
LOGBACK-1248 Support minimum length in %i filename pattern

johngmyers:


blling:
How to specify the `i` start from 1(it is zero now)?

PR #379, opened 2017-06-23
LOGBACK-1317 Change the scope of the servlet-api from compile/optional to provided

ebourg:
Hi,

The dependency for the Servlet API has the `compile` scope and is marked as optional. Usually in this situation the `provided` scope should be used because it's supposed to be provided by the servlet container.


ceki:
Can you please add a jira issue referencing this PR?

ebourg:
done: https://jira.qos.ch/browse/LOGBACK-1317

ebourg:
Thanks!

PR #391, opened 2017-08-17
Added support for Half-Daily log rotation

Skipua:
I've found out that logback is not doing half-daily log rotation when 'yyyy-MM-dd-a' pattern is given, even tough it has such PeriodicityType


logback-core/src/main/java/ch/qos/logback/core/rolling/helper/PeriodicityType.java
```diff
@@ -19,7 +19,13 @@
 
     // The followed list consists of valid periodicy types in increasing period
     // lengths
-    static PeriodicityType[] VALID_ORDERED_LIST = new PeriodicityType[] { TOP_OF_MILLISECOND, PeriodicityType.TOP_OF_SECOND, PeriodicityType.TOP_OF_MINUTE,
-            PeriodicityType.TOP_OF_HOUR, PeriodicityType.TOP_OF_DAY, PeriodicityType.TOP_OF_WEEK, PeriodicityType.TOP_OF_MONTH };
+    static PeriodicityType[] VALID_ORDERED_LIST = new PeriodicityType[]{TOP_OF_MILLISECOND,
+            TOP_OF_SECOND,
+            TOP_OF_MINUTE,
+            TOP_OF_HOUR,
+            HALF_DAY,
```
Skipua: Added HALF_DAY to the list

logback-core/src/test/java/ch/qos/logback/core/rolling/helper/RollingCalendarTest.java
```diff
@@ -30,7 +31,8 @@
 
 public class RollingCalendarTest {
 
-  String dailyPattern = "yyyy-MM-dd";
+   String dailyPattern = "yyyy-MM-dd";
+   private static final String ISO_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSS zzz";
```
michael-o: this  isn't ISO format. It has to be XXX.

Skipua: Changed it just to "DATE_FORMAT" to avoid any confusions. It's only used in test to go from string to date and back

PR #399, opened 2017-11-29
Handle circular exceptions, in relationship to LOGBACK-1027

NMcCloud:


FabianFrank:
I just ran into this in combination with AmazonKinesis, would be great to see this fix merged.

joakime:
Eclipse Jetty ran into this with SSL/TLS handshake_exceptions on OpenJDK 8u252 and OpenJDK 11.0.7

Circular Exception references are legitimate and even Java's own `Throwable.printStackTrace()` supports them properly.
It sucks to have to write a TurboFilter to work around this bug in logback.

The workaround (in case anyone else sees this issue)

``` java
package org.eclipse.jetty.logbackhack;

import java.io.IOException;
import java.io.PrintWriter;
import java.io.StringWriter;
import java.util.HashSet;
import java.util.Set;

import ch.qos.logback.classic.Level;
import ch.qos.logback.classic.Logger;
import ch.qos.logback.classic.turbo.TurboFilter;
import ch.qos.logback.core.spi.FilterReply;
import org.slf4j.Marker;

public class RecursiveThrowableFixTurboFilter extends TurboFilter
{
    public static void init()
    {
        ch.qos.logback.classic.Logger rootLogger = (ch.qos.logback.classic.Logger)LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
        rootLogger.getLoggerContext().addTurboFilter(new RecursiveThrowableFixTurboFilter());
    }

    @Override
    public FilterReply decide(Marker marker, Logger logger, Level level, String format, Object[] params, Throwable cause)
    {
        if (hasRecursiveThrowableReference(cause, null))
        {
            int locationAwareLoggerInteger = Level.toLocationAwareLoggerInteger(level);
            logger.log(marker, logger.getName(), locationAwareLoggerInteger, format, params, new SafeException(cause));
            return FilterReply.DENY;
        }

        return FilterReply.NEUTRAL;
    }

    private boolean hasRecursiveThrowableReference(Throwable cause, Set<Throwable> seen)
    {
        if (cause == null)
            return false;

        if (seen == null)
        {
            seen = new HashSet<>();
        }
        else if (!seen.add(cause))
        {
            return true;
        }

        for (Throwable suppressed : cause.getSuppressed())
        {
            if (hasRecursiveThrowableReference(suppressed, seen))
            {
                return true;
            }
        }
        return hasRecursiveThrowableReference(cause.getCause(), seen);
    }

    public static String toStackTrace(Throwable cause)
    {
        try (StringWriter sw = new StringWriter();
             PrintWriter pw = new PrintWriter(sw))
        {
            cause.printStackTrace(pw);
            return sw.toString();
        }
        catch (IOException e)
        {
            return "Unable to produce Stacktrace for " + cause;
        }
    }

    public static class SafeException extends RuntimeException
    {
        public SafeException(Throwable cause)
        {
            super(toStackTrace(cause));
        }
    }
}
```

Just call `RecursiveThrowableFixTurboFilter.init();` VERY early in your program, before you have any Loggers created.

logback-classic/src/main/java/ch/qos/logback/classic/spi/ThrowableProxyUtil.java
```diff
@@ -179,6 +187,11 @@ public static void subjoinFirstLineRootCauseFirst(StringBuilder buf, IThrowableP
     }
 
     private static void subjoinExceptionMessage(StringBuilder buf, IThrowableProxy tp) {
-        buf.append(tp.getClassName()).append(": ").append(tp.getMessage());
+        //The JVM does not print the message if it is null
```
baltiyskiy, line 190: How is this related to the main purpose of this PR? Are you fixing some other issue here as well?

baltiyskiy:
Nice PR! However, some issues are still not addressed --- there are IThrowableProxy graph traversals that can similarly loop infinitely:
 - `ch.qos.logback.classic.pattern.RootCauseFirstThrowableProxyConverter.recursiveAppendRootCauseFirst`,
 - `ch.qos.logback.classic.pattern.ThrowableProxyConverter.recursiveAppend()`,
 - `ch.qos.logback.classic.db.DBAppender.insertThrowableDBAppender.insertThrowable()`,
 - `ch.qos.logback.classic.html.DefaultThrowableRenderer.render()`,
 - `ch.qos.logback.classic.net.SyslogAppender.postProcess()`.

 The latter 3 are in fact only cause-chain traversals. Potentially there can be cause-chain only cycles as well. (That these appenders do not support "suppressed" makes me wonder if they need to be updated to print suppressed exceptions as well.)

logback-classic/src/main/java/ch/qos/logback/classic/spi/ThrowableProxyUtil.java
NMcCloud: You are right, it has nothing to do with the main purpose of the PR.  I saw another issue as I created the PR related to a test case where when the message was null, logback would print "null".  The JRE would not print "null", and looking at the test cases, the intent of logback with regard to exceptions was to mimic what the JRE did.

PR #400, opened 2017-11-29
Suppress duplicate messages for a while

jackhammer2k:
Currently the DuplicateMessageFilter filters messages forever which is not optimal. I believe in most cases people want to filter annoying in fixed-interval reoccurring messages or errors temporarily occurring in high volume. In the latter case you definitely want to "know" when it occurs again (actually you're also interested when its over, but that's a bit more difficult).
Thus the DuplicateMessageFilter shall log messages again after they did not occur for a while.

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/turbo/DuplicateMessageFilter.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/turbo/DuplicateMessageFilter.java
@@ -37,15 +37,20 @@ public class DuplicateMessageFilter extends TurboFilter {
      * The default number of allows repetitions.
      */
     public static final int DEFAULT_ALLOWED_REPETITIONS = 5;
+    /**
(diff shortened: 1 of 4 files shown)


ceki:
@jackhammer2k Expiring messages by time is an interesting idea. However, it may not be suitable for everyone.I rather it were an option. Deriving a new class named say "TimedDuplicateMessageFilter" using  ExpiringLRUMessageCache is one possibility. 

Two other minor issues,

1) duration for the expiry should be specified with ch.qos.logback.core.util.Duration and not long (milliseconds).

2) for testing purposes time should be artificially settable. See for example TimeBasedFileNamingAndTriggeringPolicyBase.artificialCurrentTime

Waiting 50 ms for a test to complete is a no go!

jackhammer2k:
Allright, currently I'm quite busy. When I'm free again I will change that.

PR #403, opened 2017-12-07
LOGBACK-1356 Append arguments to message when all are unused.

epabst:
This intentionally doesn't depend on the fix for SLF4J-422
so that logback stays compatible with prior versions of slf4j-api.

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java
@@ -290,12 +290,32 @@ public class LoggingEvent implements ILoggingEvent {
             return formattedMessage;
         }
         if (argumentArray != null) {
-            formattedMessage = MessageFormatter.arrayFormat(message, argumentArray).getMessage();
+            formattedMessage = formatMessage(message, argumentArray);
         } else {
             formattedMessage = message;
         }
 
         return formattedMessage;
+    }
+
+    static String formatMessage(String message, Object[] argumentArray) {
+        String result;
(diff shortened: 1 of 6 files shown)


epabst:
The PR build failure is unrelated:
[ERROR] /home/travis/build/qos-ch/logback/logback-classic/src/main/java/module-info.java:[8,21] module not found: groovy

huxi:
I'm not a fan of this change. The [InvalidSlf4jMessageFormat PMD rule](https://pmd.github.io/latest/pmd-java/rules/java/logging-java.html#InvalidSlf4jMessageFormat) would also barf.

epabst:
@huxi Do you have a recommendation on the problem this was intended to solve?  See https://jira.qos.ch/browse/LOGBACK-1356

huxi:
Not really.

It looks like Logstash is using the (AFAIK) undocumented feature that superfluous arguments aren't dropped by Logback but are instead left in the event.

I once suggested that this should be changed/fixed, dropping any unused arguments beside the optionally available `Throwable`. My rationale for this was that those additional arguments only use up space in case of serialized events without serving any purpose.

But it seems that what I once considered a bug or shortcoming has now grown into being a feature used by third parties...

The positive thing is that I don't decide about changes like this. I just wanted to add my 2 cents.

If this PR is accepted then someone should tell the PMD devs so they can support/ignore this special case in their rule.

epabst:
That makes sense to update PMD.  The fewer PMD issues to tell users about the better, by having the code handle cases directly.

epabst:
https://github.com/qos-ch/slf4j/pull/191 will provide a mechanism for a more thorough fix in the future.

PR #404, opened 2017-12-22
LOGBACK-543 Use Throwable.toString

bjkail:
#### Steps to reproduce
Log an exception with an overridden `toString`.  For example:
```java
public class TestException extends RuntimeException {
  public String toString() {
    return super.toString() + " [extra state]";
  }
}
```

Actual:
```
test.TestException
	at test.Main.main(Main.java:3)
```

Expected:
```
test.TestException [extra state]
	at test.Main.main(Main.java:3)
```

#### Fix
Add a new `ThrowableProxy.string` field and `getString` method.  The field is optional, and it is set to null if it is equal to the default `Throwable.toString` implementation with class name and message, which avoids overhead in `ThrowableProxyVO` for the common case.


johannesherr:
Wow, that was quick!

I have checked out the current master and ran our test suite with it. I can confirm that the log message now contains the expected additional details.

Thank you!

PR #409, opened 2018-03-10
LOGBACK-1219 changes in logback-access to allow S/sessionID usage in Jetty 9.3+ 

ledstellar:
When Session ID field is used in logback-access it tries to create new HttpSession instead of just probe/discover it. That happens in final stage of HTTP request processing and result in "IllegalStateException: Response is committed".

In this fix getSession() call replaced by getSession(false) as proposed in issue to avoid new session creation.

Tested with jetty-9.4.8.v20171121


VirusAlex:
still not fixed

ceki:
Fixed in https://github.com/qos-ch/logback/commit/e84a52d

PR #416, opened 2018-05-01
small refactor in OutputStreamAppender

nadavwe:
Looks like this code has been duplicated in the recent refactor.
this also makes it possible to override the output mechanism - as the `writeOut` method is protected while the `writeBytes` method is private.


nadavwe:
@ceki

nadavwe:
Hi @ceki, Thanks for the approval!
will you also merge this?
Thanks,
Nadav

nadavwe:
@ceki ping? :)

PR #418, opened 2018-06-04
MDC Inheritance flag (LOGBACK-624)

liranms:
The ability to choose between MDC thread inheritance it a missing feature in logback in my opinion, I've read the discussion on LOGBACK-624, and I understand why some think that this feature is bad practice, but existing products that used logback < 1.1.4 and now want to upgrade may get broken due to this change.
Also, I think its Ok if the default is set to not inherit, but there must be some way to allow changing this behavior via some flag.
I've made minimal changes that were proposed as an attachment on LOGBACK-624 by the reporter Alexander Treptow.


usatenko:
guys, please merge this functionality, it is very convenient

PR #419, opened 2018-06-11
upgrade Maven plugins

sullis:
reference: "zip slip vulnerability"

https://maven.apache.org/security-plexus-archiver.html


sullis:
@ceki

PR #421, opened 2018-06-26
Fix readme titles

timolehto:
Titles are broken:

readme_looks_like_this_now

They should probably look like this:

readme_should_probably_look_like_this


francescopellegrini:
Duplicate of #377.

PR #423, opened 2018-08-13
MsSQL - Fix event_id maximum precision to 38

okopr:
Hi,

for MsSQL is maximum precision 38 not 40. 

Database error: 

> SQL Error [2750] [S0001]: Column or parameter #15: Specified column precision 40 is greater than the maximum precision of 38

Documentation about precision: https://docs.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-2017

Thanks for merge, just minor fix.


bartfastiel:
The corresponding JIRA ticket is https://jira.qos.ch/browse/LOGBACK-792 .

PR #432, opened 2018-09-29
Compress elapsed period log files on startup

skyding228:


skyding228:
JIRA case: [LOGBACK-914](https://jira.qos.ch/browse/LOGBACK-914)

PR #439, opened 2018-11-07
If casting ‘diff’ from 'long' to 'int' forcibly, it may overflow.

chienlungcheung:
In my production environment, I found that logback cannot delete files out of date when the application is started up.

Finnaly, I find the code at [this line](https://github.com/qos-ch/logback/blob/4dd08e68321548cd85251ec389f5a832505e1268/logback-core/src/main/java/ch/qos/logback/core/rolling/helper/RollingCalendar.java#L199) which casts ‘diff’ from 'long' to 'int' forcibly will overflow when 'periodicityType' is 'TOP_OF_HOUR' and 'cleanHistoryOnStart' property is set to true.


ceki:
Duplicate. Already fixed in https://github.com/qos-ch/logback/commit/b35967934

PR #442, opened 2018-12-05
Fix addBasicProperty so it properly supports complex types

slessard:
I found a bug in logback-core/Joran where it wouldn't call the `addValue(Value v)` property on my custom Layout class. The reason for this failure was that Joran, after converting the string into an instance of `Value` would then pass the original string as the sole argument for the reflection call to `addValue(Value v)`


slessard:
Jira ticket has been created. https://jira.qos.ch/browse/LOGBACK-1445

slessard:
Is there anything I can do to help get this PR merged to master?

ceki:
My apologies for the delay, Merged with tests in commit 979d76f3f2847

PR #444, opened 2018-12-06
Support custom appender properties containing dashes

neon-dev:
Currently custom appender properties cannot be mapped from xml if they contain dashes, even though they are valid xml. With this change dashes are now supported.
Example:
```xml
<appender name="broken" class="some.custom.Appender">
    <layout>
        <pattern>%d{HH:mm:ss} [%thread] [%-5level] %logger{36} - %msg%n%ex</pattern>
    </layout>
    <property-with-many-many-many-dashes>unmapped value</property-with-many-many-many-dashes>
</appender
```
`property-with-many-many-many-dashes` will map to `public void setProperty_with_many_many_many_dashes(String value)` in the appender class.

Since the change affects more than just the `BeanDescription` mapping, this may be a breaking change for some people, but dashes also were not really supported anywhere before, so it seems very unlikely to cause issues.

Additionally, the dash to underscore conversion should be mentioned here: https://github.com/qos-ch/logback-site/blob/c4af6bb20bf1e0ff4632a0a4334cd3fcaa6c04a7/src/site/pages/manual/appenders.html#L4675-L4681
If this PR gets rejected, maybe it's still worth considering mentioning the lack of support for mapping such properties in the docs.


neon-dev:
*push*

neon-dev:
I had a hard time adding the (newly required) signed-off note to the commit message of this old commit after merging in the latest changes.
Finally got everything rebased and force pushed, so CI should be happy now.

neon-dev:
@ceki what do you think about this feature suggestion?

ceki:
PRs with no jira issues tend to fall through the cracks.

neon-dev:
No worries. I can create one, but since I deleted the repo (was cleaning up old forks) I can't change anything about the code anymore if necessary 😕

PR #445, opened 2018-12-09
Make property strings escapable

neon-dev:
Currently it's not possible to escape strings like `${some.property}` in the logback.xml.
This is especially problematic when trying to use a replacement command with named groups in the appender pattern.
Therefore something like this will attempt to insert a property value instead of the named group: `%replace(%msg){'(?&lt;namedGroup>somestuffhere)', '${namedGroup}'}`.
I kept this example as simple (stupid) as possible for demonstration purposes since real world patterns where you need named groups often are pretty complex.

With this change, you can escape these property references via the `\` character, similar to how you escape `%` pattern keywords.
Example: `\$\{namedGroup\}`
In this first draft you need to escape all possible keywords (`$` `{` `:` `}`), but it'd be possible to implement a more advanced escaping detection so you just need a simple `\` in front of the `$`.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/subst/Tokenizer.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/subst/Tokenizer.java
@@ -114,29 +114,42 @@ public class Tokenizer {
     private void handleLiteralState(char c, List<Token> tokenList, StringBuilder stringBuilder) {
         switch (c) {
         case CoreConstants.DOLLAR:
+            if (isEscaped(stringBuilder))
+                break;
             addLiteralToken(tokenList, stringBuilder);
             stringBuilder.setLength(0);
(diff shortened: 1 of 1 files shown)


ceki:
Please create a Jira issue linking to this PR.

neon-dev:
Done: https://jira.qos.ch/browse/LOGBACK-1446

Also I suggested there to maybe change the tokenizer, so it doesn't recognize `{`, `:`, and `}` anymore if they aren't part of a valid property reference preceded by a `$`. This would make a more complex escaping strategy obsolete to enable escaping with a single `\` preceding the `$`. The `isEscaped()` calls for the other three chars therefore wouldn't be needed anymore.

neon-dev:
*push*

PR #450, opened 2019-02-07
ConcurrentModificationException may be thrown when calling getCopyOfContextMap 

liranms:
ConcurrentModificationException may be thrown when calling LogbackMDCAdapter.getCopyOfContextMap when iterating over the thread local map, synchronizing over it during iteration would solve the problem.

it can happen when an external thread gets a reference of this map (even though this is thread local), e.g for debugging purposes, and modifies it


ceki:
the whole point of passing a copy is to avoid synchronisation and the threadlocal1 thing. Not going to add synchronisation.

PR #392, opened 2017-09-01
README: fix formatting of section headers

frankschmitt:
Add spaces between # and the section headers to ensure they are displayed correctly on Github

PR #452, opened 2019-03-01
Fix markup in README

ivoanjo:
Hello there! Quick fix, as otherwise it doesn't look correct on github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 
-#About logback
+# About logback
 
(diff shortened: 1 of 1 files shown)


frankschmitt:
It's funny that the 4th (?) duplicate pull request finally got through while #421, #392 and #377 all were completely ignored by the project maintainers :-/

Well, better late than never!

PR #453, opened 2019-03-14
LOGBACK-1000 MDC default value is not parsed

rhoffman12:
https://jira.qos.ch/browse/LOGBACK-1000

Pattern conversion words used as the default value of an MDC conversion word are detected and converted. As an example, the pattern `%X{akkaSource:-%logger}` will try to use the MDC context for key `akkaSource` and will fall back on the built-in `%logger` conversion if no such MDC exists.

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/pattern/MDCConverter.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/pattern/MDCConverter.java
@@ -13,23 +13,31 @@
  */
 package ch.qos.logback.classic.pattern;
 
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.PatternLayout;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


Albertmoiseev:
Why this has not been merged yet?

qixiaobo:
LGTM

logback-classic/src/main/java/ch/qos/logback/classic/pattern/MDCConverter.java
```diff
@@ -56,7 +67,7 @@ public String convert(ILoggingEvent event) {
             if (value != null) {
```
pirateskipper, line 67: `return value == null ? defaultValue.doLayout(event) : value;`

ceki:
My apologies for the lack of reaction to this issue.

However, I don't think the added complexity is worth it.

rhoffman12:
Fair enough, thanks for the feedback!

pro100jekon:
Not quite sure, why this feature is not worth. It is not technically impossible and there are cases when it might come in handy. Why?

PR #454, opened 2019-04-14
count should be synchronized

senwa:
I think the count attribute should be synchronized


Albertmoiseev:
is repo alive? why nothing is being reviewed, merged?

ceki:
There has been a very significant rewrite of the Joran framework and has taken a lot of bandwidth.

PR #462, opened 2019-07-23
Avoid synchronized blocks when copying MDC to a function-local map

efdknittlfrank:
Only wrap the new HashMap in a synchronizedMap after it has been
initialized and copied. This avoids having to enter a synchronized block
for the putAll operation.


efdknittlfrank:
This should give a little speedup if you are using e.g. [`TransferableMdc`](https://github.com/qos-ch/slf4j/pull/150).

I'd be happy if someone can benchmark, whether this really gives a little speed improvement or if the JVM is already optimizing this anyway.

ceki:
Kindly create a JIRA issue referencing this PR.

PR #477, opened 2020-03-12
LOGBACK-1509 - allocate less from Abbreviators

mkeskells:


logback-classic/src/main/java/ch/qos/logback/classic/pattern/NamedConverter.java
```diff
@@ -50,7 +55,7 @@ public String convert(ILoggingEvent event) {
         if (abbreviator == null) {
```
pirateskipper, line 55: `return abbreviator == null ? fqn : knownAbbreviations.computeIfAbsent(fqn, s -> abbreviator.abbreviate(s));`

ceki: The code as it exists in the repo is quite different now and this change is no longer applicable.

PR #484, opened 2020-03-31
Maven Reproducible builds

dialaya:
Add new build properties and update maven jar and maven source plugins version.

Context of the change: 
Make logback builds reproducible regarding the [Guide to Configuring for Reproducible Builds](https://maven.apache.org/guides/mini/guide-reproducible-builds.html)

The work is done with help of @hboutemy to achieve the goal.

### DIFF
--- a/logback-access/pom.xml
+++ b/logback-access/pom.xml
@@ -137,6 +137,7 @@
             </Import-Package>
             <Bundle-RequiredExecutionEnvironment>JavaSE-1.6
             </Bundle-RequiredExecutionEnvironment>
+            <_removeheaders>Bnd-LastModified</_removeheaders>
           </instructions>
         </configuration>
       </plugin>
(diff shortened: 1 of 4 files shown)


ceki:
We would need to add a timestamp manually which does not seem like a good idea.

hboutemy:
Hi Ceki,
or you just let the original timestamp if you don't really care its value (some tools just put the value to 1970-01-01 00:00...)
but given a jar file is a zip file, reproducibility requires to have a strategy to define a reproducible timestamp, we have no choice
notice: maven-release-plugin 3.0.0-M1 is able to update the value in pom.xml during release, but I see you don't use it

zeldigas:
Take a look at https://zlika.github.io/reproducible-build-maven-plugin/ it greatly simplifies creation of reproducible builds

hboutemy:
zlika just chose a fixed timestamp value

zeldigas:
Maybe, but this plugin is a condensed set of practices of getting reproducible artifact, that is way more reliable than random manual fixes of manifest (as an example - it supports fixing file attributes inside jar)

But more general question is more interesting - is logback still alive? Looking at amount of pull requests, I doubt.

PR #501, opened 2020-11-24
module-info.java missing dependency on java.logging

cowwoc:
class `ch.qos.logback.classic.jul.JULHelper` requires `java.logging`.

Relevant stack trace is:

```
Exception in thread "main" java.lang.IllegalAccessError: class ch.qos.logback.classic.jul.JULHelper (in module ch.qos.logback.classic) cannot access class java.util.logging.Logger (in module java.logging) because module ch.qos.logback.classic does not read module java.logging
	at ch.qos.logback.classic/ch.qos.logback.classic.jul.JULHelper.asJULLogger(JULHelper.java:66)
	at ch.qos.logback.classic/ch.qos.logback.classic.jul.JULHelper.asJULLogger(JULHelper.java:70)
	at ch.qos.logback.classic/ch.qos.logback.classic.jul.LevelChangePropagator.propagate(LevelChangePropagator.java:61)
	at ch.qos.logback.classic/ch.qos.logback.classic.jul.LevelChangePropagator.propagateExistingLoggerLevels(LevelChangePropagator.java:88)
	at ch.qos.logback.classic/ch.qos.logback.classic.jul.LevelChangePropagator.start(LevelChangePropagator.java:97)
	at ch.qos.logback.classic/ch.qos.logback.classic.joran.action.LoggerContextListenerAction.end(LoggerContextListenerAction.java:70)
	at ch.qos.logback.core/ch.qos.logback.core.joran.spi.Interpreter.callEndAction(Interpreter.java:309)
	at ch.qos.logback.core/ch.qos.logback.core.joran.spi.Interpreter.endElement(Interpreter.java:193)
	at ch.qos.logback.core/ch.qos.logback.core.joran.spi.Interpreter.endElement(Interpreter.java:179)
	at ch.qos.logback.core/ch.qos.logback.core.joran.spi.EventPlayer.play(EventPlayer.java:62)
	at ch.qos.logback.core/ch.qos.logback.core.joran.GenericConfigurator.doConfigure(GenericConfigurator.java:165)
	at ch.qos.logback.core/ch.qos.logback.core.joran.GenericConfigurator.doConfigure(GenericConfigurator.java:152)
	at ch.qos.logback.core/ch.qos.logback.core.joran.GenericConfigurator.doConfigure(GenericConfigurator.java:110)
	at ch.qos.logback.core/ch.qos.logback.core.joran.GenericConfigurator.doConfigure(GenericConfigurator.java:53)
	at ch.qos.logback.classic/ch.qos.logback.classic.util.ContextInitializer.configureByResource(ContextInitializer.java:82)
	at ch.qos.logback.classic/ch.qos.logback.classic.util.ContextInitializer.autoConfig(ContextInitializer.java:157)
	at ch.qos.logback.classic/ch.qos.logback.classic.spi.LogbackServiceProvider.initializeLoggerContext(LogbackServiceProvider.java:49)
	at ch.qos.logback.classic/ch.qos.logback.classic.spi.LogbackServiceProvider.initialize(LogbackServiceProvider.java:40)
	at org.slf4j/org.slf4j.LoggerFactory.bind(LoggerFactory.java:152)
	at org.slf4j/org.slf4j.LoggerFactory.performInitialization(LoggerFactory.java:139)
	at org.slf4j/org.slf4j.LoggerFactory.getProvider(LoggerFactory.java:418)
	at org.slf4j/org.slf4j.LoggerFactory.getILoggerFactory(LoggerFactory.java:404)
	at org.slf4j/org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:353)
```

### DIFF
--- a/logback-classic/src/main/java9/module-info.java
+++ b/logback-classic/src/main/java9/module-info.java
@@ -1,4 +1,5 @@
 module ch.qos.logback.classic { 
+  requires java.logging;
   requires org.slf4j;
   requires static java.management;
   requires static javax.servlet.api;


namannigam:
for those reaching out here, a temporary solution unless this gets merged and released would be to use `--add-reads ch.qos.logback.classic=java.logging` as an additional VM argument.

logback-classic/src/main/java9/module-info.java
```diff
@@ -1,4 +1,5 @@
 module ch.qos.logback.classic { 
+  requires java.logging;
```
iajn, line 2: This forces the module requirement for everybody using Logback Classic in a modular environment. This statement should be requires static instead.

cowwoc: @iajn Please help me understand your concern.

1. `LevelChangePropagator` is part of the logback classic JAR file. It isn't an optional plugin like the JUL to SLF4J bridge.
2. `requires` means that any code that depends on logback classic will have to provide `java.logging` on the modulepath, which coincidentally doesn't impact the end-user at all since this is guaranteed to ship with the JDK.
3. While it is true that not everyone who uses logback classic will choose to use `LevelChangePropagator` if you decide to use `requires static` every time this happens you will quickly find yourself requiring nothing. As far as I understand it, this isn't what `requires static` was meant for.
4. And finally, `requires` is not `requires transitive` so we're not saying that anyone who depends on logback classic will propagate `requires java.logging` to its own users.

Am I missing anything?

iajn: I was mostly concerned for situations where the end user uses jlink to shave off modules that are not required to slimline the JDK delivered with the application. When used with a full JDK that always includes the java.logging module I don't think it matters whether requires or requires static is used.

I'm not entirely sure what you mean with "requiring nothing", but to [quote this blog post](https://nipafx.dev/java-modules-optional-dependencies/):
_The module system allows optional dependencies with the requires static clause. Module required this way are accessible at compile time but can be absent at run time._

To me this sounds like a good use case. If the end user doesn't use the LevelChangePropagator, they also don't require the java.logging module, so it can be absent from the jlink'd JDK. If the LevelChangePropagator is used, the user needs to add the requirement for java.logging themselves.

Because Logback is usually configured in xml instead of code, this does add a minor annoyance of needing to add java.logging to the module-info (or to the --add-modules list) manually without there being any indication in the code to do so. The application should work fine on a full JDK without the explicit requirement, but would probably encounter a runtime error when using a jlink'd JDK. This requirement could of course be explicitly stated in the documentation that covers the LevelChangePropagator.

There are minor cons for both approaches. In my opinion it would be better to not force the extra module (even though it is very small) for everybody. But this is not a hill I wish to die on.

ceki: Thank you both for your thoughtful comments which are highly appreciated. I think the Jlink argument brings a new perspective which many of us  previously did not pay attention to.

ceki:
This issue was solved in logback version 1.3.4/1.4.4 in 2022-10-07 by adding `requires static java.logging` in logback-classic/../module-info.java. Note that if you installed `LevelChangePropagator`, your application presumably calls `java.util.logging` which requires `'requires java.logging`. 

Regarding jlink, it is not clear whether `requires static java.logging` forces  the inclusion of java.util.logging into the generated binary. I do can't test this in the short term but the question is somewhat interesting.

ceki:
Related to [LOGBACK-1578](https://jira.qos.ch/browse/LOGBACK-1578)

PR #506, opened 2021-02-04
ContextNameModelHandler.java  There is a fatal error

gongxuanzhang:
in ContextNameModelHandler line 24 have a "fix me"


514190950:
I don't know what the writer means
so I comment it

PR #510, opened 2021-02-14
LOGBACK-888: Add possibility to mark property file/resource as optional

jckosta:
Support attribute "optional" in property element.

Similar to the already existing support in the include element, it
prevents errors loading non-existent property files/resources.


jckosta:
This PR applies to the 1.2.x branch. PR #511 does the same for the master branch.

PR #515, opened 2021-03-05
Avoid StackOverflowError (fixes LOGBACK-1454)

jpstotz:
by Exceptions that build loops via nested Exceptions in getCause()/getSuppressed()

I also removed some code only required for Java 6 and older which is no longer supported.  

Fixes https://jira.qos.ch/browse/LOGBACK-1454

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/spi/ThrowableProxy.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/spi/ThrowableProxy.java
@@ -15,8 +15,11 @@ package ch.qos.logback.classic.spi;
 
 import ch.qos.logback.core.CoreConstants;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.IdentityHashMap;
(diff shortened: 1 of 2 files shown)


jpstotz:
@ceki Don't you think it would be a good idea to process the `fix me` text that has been added in 2019 which breaks compilation? 
https://github.com/qos-ch/logback/blame/master/logback-classic/src/main/java/ch/qos/logback/classic/model/processor/ContextNameModelHandler.java#L25

Having a continuous integration only makes sense if the code is compliable, do you agree?

lukasz20p (approved):

clamothe:
Thanks for merging this. It would be great to have it in a release soon.

wpanas:
Na razie zmiana jest tylko w 1.3, a ta wersja nie jest kompatybilna z Spring Boot https://github.com/spring-projects/spring-boot/issues/12649

wpanas:
@ceki I wonder if this can be fixed for 1.2.x. Right now 1.3.x is unsupported for Spring Boot. What do you think? If you agree, I can prepare a pull request.

wpanas:
As far as I see, it's addressed in https://github.com/qos-ch/logback/pull/535

PR #524, opened 2021-08-04
fix LOGBACK-1566

aomsweet:
Signed-off-by: aomsweet <aomsweet@tutanota.com>


ceki:
While they identify the problem, they do not take into account that a CompositeConverter may have multiple children.

PR #525, opened 2021-08-04
fix LOGBACK-1566

aomsweet:
Signed-off-by: aomsweet <aomsweet@tutanota.com>


ceki:
While they identify the problem, they do not take into account that a CompositeConverter may have multiple children.

aomsweet:
@ceki  Thanks for your review☕. I did ignore the possible children of CompositeConverter😅.

aomsweet:
@ceki The 1.2.x version has the same problem, but you only fixed 1.3.x.

PR #529, opened 2021-09-10
moving to jakarta.servlet: this service provider should also be renamed

cocorossello:
This one should be renamed.

I think that (at least) javax.mail and javax.activation references are also incorrect, but I suppose it's an ongoing work


ceki:
Sorry. Saw this PR too late.

ceki:
Many thanks for catching this.

PR #532, opened 2021-10-14
Fixing behavior of logback-access in the modern era

joakime:
+ Fixes [LOGBACK-1580](https://jira.qos.ch/browse/LOGBACK-1580) to report the proper `HttpServletResponse.getStatus()` that was used when the response was committed.
+ Works around [LOGBACK-1581](https://jira.qos.ch/browse/LOGBACK-1581) in `AccessEvent` to not fail logging when the `HttpServletRequest.getParameterNames()` or `HttpServletRequest.getParameter(String)` methods are used against unread or even bad  request objects.
+ Updated Javadoc (needs to be copied over to logback manual though)
+ Must use OpenJDK 1.8 to compile and release.
+ Still compiles to Java 1.6 bytecode for release.

### DIFF
--- a/logback-access/src/main/java/ch/qos/logback/access/jetty/RequestLogImpl.java
+++ b/logback-access/src/main/java/ch/qos/logback/access/jetty/RequestLogImpl.java
@@ -41,7 +41,6 @@ import ch.qos.logback.core.util.StatusPrinter;
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.RequestLog;
 import org.eclipse.jetty.server.Response;
-import org.eclipse.jetty.util.Jetty;
 import org.eclipse.jetty.util.component.LifeCycle;
 
 /**
@@ -222,31 +221,30 @@ public class RequestLogImpl extends ContextBase implements RequestLog, LifeCycle
     String fileName;
     String resource;
 
-    // Default to non-modern Jetty (ie: Jetty 9.3 or earlier)
     // Jetty 9.4.x and newer is considered modern.
-    boolean modernJettyRequestLog = false;
+    boolean modernJettyRequestLog;
     boolean quiet = false;
 
     public RequestLogImpl() {
         putObject(CoreConstants.EVALUATOR_MAP, new HashMap<String, EventEvaluator<?>>());
 
         // plumb the depths of Jetty and the environment ...
-        try {
-            Class.forName("jakarta.servlet.http.HttpServlet");
+        if (classIsPresent("jakarta.servlet.http.HttpServlet")) {
             throw new RuntimeException("The new jakarta.servlet classes are not supported by this " +
                 "version of logback-access (check for a newer version of logback-access that " +
                 "does support it)");
-        } catch(Throwable ignore) {
-            // Class doesn't exist, assumption is that this is normal javax.servlet environment.
         }
 
         // look for modern approach to RequestLog
+        modernJettyRequestLog = classIsPresent("org.eclipse.jetty.server.CustomRequestLog");
+    }
+
+    private boolean classIsPresent(String className) {
         try {
-            Class.forName("org.eclipse.jetty.server.CustomRequestLog");
-            // Class exists, treat as modern Jetty.
-            modernJettyRequestLog = true;
-        } catch(Throwable ignore) {
-            // Class doesn't exist, this is an old school Jetty.
+            Class.forName(className);
+            return true;
+        } catch (ClassNotFoundException e)  {
+            return false;
         }
     }
 


logback-access/src/main/java/ch/qos/logback/access/jetty/RequestLogImpl.java
```diff
@@ -112,32 +209,66 @@
...
+
+        // plumb the depths of Jetty and the environment ...
+        try {
+            Class.forName("jakarta.servlet.http.HttpServlet");
+            throw new RuntimeException("The new jakarta.servlet classes are not supported by this " +
+                "version of logback-access (check for a newer version of logback-access that " +
+                "does support it)");
+        } catch(Throwable ignore) {
+            // Class doesn't exist, assumption is that this is normal javax.servlet environment.
+        }
```
dimas, line 241: This `catch` will swallow the `Exception` you are trying to raise here...

logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java
```diff
@@ -335,13 +335,22 @@ public void buildRequestHeaderMap() {
...
+        try {
+            Enumeration<String> e = httpRequest.getParameterNames();
+            if (e == null) {
+                return;
+            }
+            while (e.hasMoreElements()) {
+                String key = e.nextElement();
+                requestParameterMap.put(key, httpRequest.getParameterValues(key));
+            }
+        } catch(Throwable t) {
```
joakime, line 347: The Servlet Spec doesn't specify the exceptions that can happen here.
It can be something as simple as an IOException (which is a checked exception) that has to be wrapped in something else.
It can be a parsing issue, it can be a Socket issue, it can be charset issue, etc...

Unlike a normal webapp and how it handles failure, a RequestLog cannot skip its logging.
Ideally, a RequestLog should never call a method that changes the Request or Response state.

logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java
```diff
@@ -500,9 +509,14 @@ public String getRequestContent() {
                         buf.append("");
                     }
                 }
-            } catch (Exception e) {
-                // FIXME Why is try/catch required?
-                e.printStackTrace();
+            } catch (Throwable t) {
```
joakime, line 512: Ditto here.

The Servlet Spec doesn't specify the exceptions that can happen here.
It can be something as simple as an IOException (which is a checked exception) that has to be wrapped in something else.
It can be a parsing issue, it can be a Socket issue, it can be charset issue, etc...

Unlike a normal webapp and how it handles failure, a RequestLog cannot skip its logging.
Ideally, a RequestLog should never call a method that changes the Request or Response state.

logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java
joakime: Here's a few things that can happen at the `Error` level within the Servlet that are non-fatal for the JVM or ServletContext ...

* XML initialization errors (stream, parser, validation, etc) (endpoint specific)
* java.io.IOError with non-socket based connections/connectors (like UnixSocket)
* ServiceLoader and Service provider errors. (endpoint specific)
* java.nio.Charset encode/decode errors. (request specific)

joakime: The logging here in RequestLog should not fail for a fault out of its control.

zUniQueX:
@ceki could we get this merged? This prevents updating Jetty to 10.x.x, since logback assumes a method of signature `org.eclipse.jetty.http.HttpFields org.eclipse.jetty.server.Response.getHttpFields()` instead of `org.eclipse.jetty.http.HttpFields$Mutable org.eclipse.jetty.server.Response.getHttpFields()` to be provided.

logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java
zUniQueX: @joakime is this implementation still correct? I've tested Logback 1.3.0-alpha14 (which basically uses the same implementation without the try block) with Jetty 11.0.8 and at this method a `NullPointerException` gets thrown from [here](https://github.com/eclipse/jetty.project/blob/jetty-11.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java#L478).

zUniQueX:
Any updates on this? This still prevents using recent Jetty versions in combination with Logback

zUniQueX:
@ceki After revisiting this, I think this PR is the only solution to use Jetty 10.0.x with Logback.

As mentioned in [[Logback-1610]](https://jira.qos.ch/browse/LOGBACK-1610) the correct approach to support Jetty 10.0.x would be to use different modules for Jetty 10.0.x and Jetty 11.0.x. But this isn't possible at the moment (see 2ff2d4ef398b5c6863f43b8dc8e1d741c6d840d3). `IAccessEvent` would have to be provided in two versions and therefore `logback-access` would completely have to be split into two modules. Additionally, it would require a second module for `logback-core`, because the `ViewStatusMessagesServletBase` also depends on classes from the `servlet-api` package.

So my suggestion would be to provide support for Jetty 10.0.x in the Logback 1.2.x branch and remove it from the 1.3.x series in favor of support for Jetty 11.0.x.

joakime:
I'm out at the moment, I can review this at the end of the month.

ceki:
Merged into master (logback 1.3.0-bata1).

PR #536, opened 2021-11-09
Bump JDK-11 & source cleanup

zorglube:


zorglube:
Sorry guys, code-style formating is comming. 

My bad.

ceki:
Hello, 
These changes are not useful and won't be merged.

zorglube:
> Hello, These changes are not useful and won't be merged.

Sad but I can understand. 
May I ask, going to JDK 11 isn't on the logbak & slf4j roadmap ?

hazendaz:
@zorglube third parties have to balance support with legacy jdks.  The current slf4j/logback works fine with jdk up through 17 as it is so there is little benefit to making it require and that would just lose customers.  Downstream you can use whatever jdk you like.  The main issues or concerns with most third parties that I can think of were lock step addressing support for latest jdks, it never meant they required them.  This is why a lot of times you will see numerous builds such as jdk 8, 11, 17, and 18-ea for projects.  That is to show evidense that they work on those revisions.

zorglube:
@hazendaz Thank for your answer.

zorglube:
Thank you for the explanation.


Le sam. 13 nov. 2021 à 04:01, Jeremy Landis ***@***.***> a
écrit :

PR #538, opened 2021-12-13
LOGBACK-1590 - Backport of "Improve support for the JANSI library"

NilsRenaud:
Here is the copied description of cherry-picked commit : 58f15b51906b8ea25a509a46b33d87aef800d446

---
Motivation:

The Jansi library provides an AnsiPrintStream that intercepts ANSI
sequences and allows the application to react to them accordingly.
Jansi is compatible with a variety of terminals, including xterm-like
terminals for Windows like Mintty and MinGW.

Jansi has a good algorithm to detect whether or not ANSI
sequences will be correctly interpreted or not, and if not, its default
behavior is to filter out such sequences.

Logback's ConsoleAppender has a withJansi option. Unfortunately,
Logback bypasses Jansi's detection algorithm and boldly
assumes that if the OS is Windows, then the terminal is a DOS prompt.
So while DOS prompts get colored, terminals like Mintty or MinGW not
only lose their coloring but are sent DOS escape sequences that they
can't interpret.

Modifications:

- Amend ConsoleAppender so that when withJansi is true, then all
OutputStreams are wrapped in an AnsiPrintStream, effectively delegating
to the JANSI library the decision to filter out ANSI escape sequences
or not, depending on the terminal capability. This also allows users
to disable or customize the coloring behavior by using system properties
recognized by JANSI, such as jansi.strip.
- Amend the documentation and stress that the JANSI library is the
recommended solution for Windows systems, as well as for cross-platform
deployments that include a mix of Windows and non-Windows systems.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
@@ -14,13 +14,14 @@
 package ch.qos.logback.core;
 
 import java.io.OutputStream;
+import java.io.PrintStream;
+import java.lang.reflect.Method;
 import java.util.Arrays;
 
 import ch.qos.logback.core.joran.spi.ConsoleTarget;
 import ch.qos.logback.core.status.Status;
 import ch.qos.logback.core.status.WarnStatus;
-import ch.qos.logback.core.util.EnvUtil;
-import ch.qos.logback.core.util.OptionHelper;
+import ch.qos.logback.core.util.Loader;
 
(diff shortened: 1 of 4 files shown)


NilsRenaud:
@ceki up ?

NilsRenaud:
I give it another round of up (@ceki )

NilsRenaud:
Periodic reminder... @ceki should I continue to hope for a merge ?

NilsRenaud:
We finally upgraded to a more recent logback version.

PR #541, opened 2021-12-17
Update Jansi library to 2.1.0

NilsRenaud:
Motivation:

Use the last version of Jansi library to benefit of its last updates/bugfixes.

Modifications:

- Amend ConsoleAppender to use the new JANSI API. Since 2.0, the "wrapSystemOut"/"wrapSystemErr" methods have been removed and now JANSI uses "FileDescriptor.out" and "FileDescriptor.err" directly. As stated in [this JANSI commit](https://github.com/fusesource/jansi/commit/57aa84d4120395922458f61f20e741198f051087) :

> This has the downside of not playing well if the user has overridden the system streams previously, but any kind of wrappers around the streams should happen after JANSI processing

So it is not possible to test the output on stdout/err anymore.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
@@ -14,7 +14,6 @@
 package ch.qos.logback.core;
 
 import java.io.OutputStream;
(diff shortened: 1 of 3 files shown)


NilsRenaud:
I'll upgrade to 2.4.0 instead.

PR #542, opened 2021-12-17
LOGBACK-1595 Update JANSI library to 2.4.0

NilsRenaud:
Motivation:

Use the last version of JANSI library to benefit of its last updates/bugfixes.

Modifications:

- Amend ConsoleAppender to use the new JANSI API. Since 2.0, the "wrapSystemOut"/"wrapSystemErr" methods have been removed and now JANSI uses "FileDescriptor.out" and "FileDescriptor.err" directly. As stated in this JANSI commit https://github.com/fusesource/jansi/commit/57aa84d4120395922458f61f20e741198f051087 :
  > This has the downside of not playing well if the user has overriden the system streams previously, but any kind of wrappers around the streams should happen after JANSI processing

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
@@ -14,7 +14,6 @@
 package ch.qos.logback.core;
 
 import java.io.OutputStream;
-import java.io.PrintStream;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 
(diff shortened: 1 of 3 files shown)


gnodet:
See https://github.com/gnodet/logback/commit/e60cd353b834f7e656345ad403f0cfed128e2863 for a fix

NilsRenaud:
Hi there (@ceki), any point of view on this PR ? Should we reject/merge it ?

logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
```diff
@@ -77,27 +75,23 @@ private void targetWarn(String val) {
 
     @Override
     public void start() {
-        OutputStream targetStream = target.getStream();
         // enable jansi only if withJansi set to true
-        if (withJansi) {
-            targetStream = wrapWithJansi(targetStream);
-        }
-        setOutputStream(targetStream);
+        setOutputStream(withJansi ? getJansiStream() : target.getStream());
```
ceki, line 79: Why change this?

NilsRenaud: Indeed, I can revert. I found the 1 liner easier to read though.

PR #544, opened 2021-12-19
Add feature timer for DuplicateMessageFilter

ni-mi:
https://jira.qos.ch/browse/LOGBACK-715 - Add feature timer for DuplicateMessageFilter.
Current DuplicateMessageFilter is unchanged, unless the CacheResetWindow property is set to anything other then Long.MaxLong.


ni-mi:
Multiple commits

PR #550, opened 2021-12-22
LOGBACK-1601 indicate users that groovy is NOT supported in this version

yujin-hong:


yujin-hong:
@ceki  
I'm sorry I forgot to assign myself to my problem.
https://jira.qos.ch/browse/LOGBACK-1601

I'd appreciate your review on this one. 
Thank you

nngo (changes requested):
if groovy configuration is no longer supported it should be removed from the message in the exception message on line 70 (i.e. to just "Should be .xml" vs "Should be either .groovy or .xml")
```java
throw new LogbackException("Unexpected filename extension of file [" + url.toString() + "]. Should be either .groovy or .xml");
```

ceki:
This issue is fixed in logback version 1.2.10 and 1.3.0-alpha12

PR #551, opened 2021-12-23
Update LoggingEvent.java

ydq:
The initialized Map can be written


ceki:
The returned map may be null according to the SLF4J API docs. Creating a new HashMap does not sound too good to me.

ydq:
> The returned map may be null according to the SLF4J API docs. Creating a new HashMap does not sound too good to me.

Thanks for your reply.  

In some special scenarios, if I want to store some variables in MDC, I may have to use reflection to modify mdcPropertyMap, but if I use new HashMap,I can put the variables directly into the MDC

PR #554, opened 2022-01-12
Introduce global ILoggingEvent wrapper for 1.3+

sokomishalov:


ceki:
Hello @sokomishalov 

Thank you for contributing this PR. Can you explain the purpose of the event-transformer? A concrete use case would be nice.

sokomishalov:
@ceki Sure!
Test that covered this functionality partially explained why such wrapping would be useful.
Let's imagine we have several appenders with different formats/layouts and we need to mask sensitive data in either event message or some other event data. Instead of implementing such masking logic in every single appender or layout, we are able to wrap an event single time with custom logic.

PR #568, opened 2022-04-28
Add missing word "find" into one log line

ZeroOne3010:
There was the word "find" missing from one log line, so I added it. I also created a unit test that covers it, for it was not covered at all before.


ceki:
@ZeroOne3010  Can you please sign you commit with the -s option?

ZeroOne3010:
@ceki I certainly can! Please see #570

PR #573, opened 2022-05-25
LOGBACK-1227 Avoid excessive filter list clone in TurboFilterList

alexlmikh:
Current implementation clones a list via ```toArray```, which results in excessive allocation in a high-load scenarios.
CoW list iterator is guaranteed to operate on a last known list snapshot already, so there is no need to clone it.


janickr:
Any idea if this PR has a chance to be merged? 
LOGBACK-1227 causes quite a lot of allocations, since it happens on every log statement (even when the log message is ignored because the log level was too low)

It's a bit unfortunate that TurboFilterList extends CopyOnWriteArrayList while at the same time it's part of the public API, because it would be better to have a different implementation altogether that doesn't even need the iterator, but it would be difficult to keep it backwards compatible.

PR #583, opened 2022-08-21
Add implementation version in MANIFEST

brenuart:
Configure the jar plugin to include the default implementation and specification entries in the MANIFEST file such that Package#getImplementationVersion() and Package#getSpecificationVersion() return meaningful values.

closes [#1597](https://jira.qos.ch/browse/LOGBACK-1597)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -284,6 +284,13 @@
           <version>${maven-jar-plugin.version}</version>
           <configuration>
             <outputTimestamp>${project.build.outputTimestamp}</outputTimestamp>
+            <archive>
+            <index>true</index>
+              <manifest>
+                <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+              </manifest>
+            </archive>
           </configuration>
         </plugin>
         <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -285,8 +285,11 @@
           <configuration>
             <outputTimestamp>${project.build.outputTimestamp}</outputTimestamp>
             <archive>
-            <index>true</index>
+              <index>true</index>
               <manifest>
+                <!-- Add version information into the MANIFEST so it is accessible by
+                     calling Package#getImplementationVersion()
+                 -->
                 <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                 <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
               </manifest>


ceki:
does it do what it is supposed to do?

brenuart:
> does it do what it is supposed to do?
Well... this leverage standard maven-jar-plugin feature. This will for instance add the following entries in the MANIFEST:
```
Specification-Title: Logback Core Module
Specification-Vendor: QOS.ch
Specification-Version: 1.3
Implementation-Title: Logback Core Module
Implementation-Vendor: QOS.ch
Implementation-Version: 1.3.0-SNAPSHOT
```

These entries are from where Package#getImplementationVersion() takes its info.

> A Unit test?
I can add one - not to test if the plugin did its job correctly but at least this test would fail if the POM is modified and the configuration altered/removed.

ceki:
I just realized that the unit test has to live in another module so it can test the state of the finished jar.

brenuart:
Hmmm... this is strange indeed. 

It appears that Package#getImplementationVersion() isn't working (as I) expected when the jar is exploded on the filesystem - which is what happens when running tests (the class path is build from the target/classes directory instead of the jar that has just been produced).
As you said, testing that calling Package#getImplementationVersion() works as expected must be done in another project... which is weird IMHO.

Do we actually need this test? Can't we just rely on the proper pom configuration (possibly adding a comment so nobody will ever remove that config) ?

brenuart:
(BTW, I have no idea why builds are failing - looks like it was already the case in main from where my branch is created...)

ceki:
> (BTW, I have no idea why builds are failing - looks like it was already the case in main from where my branch is created...)

Yah, strange.

ceki:
> Do we actually need this test? Can't we just rely on the proper pom configuration (possibly adding a comment so nobody will ever remove that config) ?

OK for adding a comment in pom.xml. 

Adding a unit test would be better. I wonder if the **ant** part of the build gets executed after jar creation. If that is the case, we can add the unit test in the ant part.

brenuart:
> Adding a unit test would be better. I wonder if the ant part of the build gets executed after jar creation. If that is the case, we can add the unit test in the ant part.

The test phase appears after the package phase - so, yes, the jar is available _before_ the tests are executed. It is therefore technically possible to invoke an ANT script (or any other kind of script like Groovy for instance) to read and assert the content of the Manifest...

A couple of remarks tho:
- should we do it for all jars (core, access, classic) ?
- is it enough to check for the presence of one of the required entries or should we also validate the value? Checking the value is similar to testing the behaviour of the jar plugin... (which is obviously not our job)
- at the end we could simply check the parent pom and make sure the required XML elements are still there...

At the end all of this seems to much and a simple comment in the pom enough... Up to you - tell me.

PR #586, opened 2022-09-02
Fix request log on Jetty 10.0.x

zUniQueX:
The compiled JAR for `logback-access:1.3.0` doesn't fully work with Jetty 10.0.x.

When decompiling the JAR with `javap`, the following `Methodref` is used in the `JettyModernServerAdapter#buildResponseHeaderMap()` method:
```
#49 = Methodref #14.#50 // org/eclipse/jetty/server/Response.getHttpFields:()Lorg/eclipse/jetty/http/HttpFields;

#50 = NameAndType #51:#52 // getHttpFields:()Lorg/eclipse/jetty/http/HttpFields;
#51 = Utf8 getHttpFields
#52 = Utf8 ()Lorg/eclipse/jetty/http/HttpFields;
```

Since the return type was changed to `HttpFields.Mutable` in Jetty 10.0.x, the method isn't found and the following exception is thrown:
```
java.lang.NoSuchMethodError: 'org.eclipse.jetty.http.HttpFields org.eclipse.jetty.server.Response.getHttpFields()'
```

However, in both Jetty versions the return types implement `Iterable<HttpField>`, so we can resolve the correct method via reflection and cast the result to `Iterable<HttpField>`.

Additionally, I think it's a good idea to make the `makeJettyServerAdapter` method `protected` to give users the flexibility to disable the reflection usage, if this gets a performance issue.

CC @joakime Maybe you can take a look at this, if this is correct or if I'm missing something.


zUniQueX:
Anything I can do to make it easier getting this merged?

PR #591, opened 2022-09-09
Make better use of assertThrows() in tests

rhowe:
Use assertThrows() to positively assert that an exception was thrown, rather than using fail() to assert that it wasn't not thrown.


zUniQueX:
I just wondered why you didn't use AssertJ. But only `logback-parent` has a dependency on it and the version is 1.7.1 :sweat_smile:

PR #598, opened 2022-09-29
fix: blocked when many exception happened

nomeatcoder:
image
When a large number of exceptions occur, the request is blocked because the class starting with sun. reflect cannot hit the cache when loading


ceki:
Thank you for your PR. Can you please create a jira issue for this PR at https://jira.qos.ch? 

Which version of logback are you using? Since logback version 1.1.4 packaging data is disabled by default.

nomeatcoder:
> Thank you for your PR. Can you please create a jira issue for this PR at https://jira.qos.ch?
> 
> Which version of logback are you using? Since logback version 1.1.4 packaging data is disabled by default.

  Since logback version 1.1.4 packaging data is disabled by default, this problem has resloved.But if it is enabled, this problem still exists.I have created a jira issue for this PR at https://jira.qos.ch/browse/LOGBACK-1697

PR #600, opened 2022-09-30
Performance fix for copy array every time

sgoryachkin:
toArray() clones the array on every call to isDebugEnabled()


ceki:
Copy-before-iterate is necessary to avoid ConcurrentModificationExceptions and such.

sgoryachkin:
IndexOutOfBoundsException was added likewise size==1 case

PR #602, opened 2022-10-07
Fix for copy array every time (second try)

sgoryachkin:
No any concurrency problem with this code because: 
- TurboFilterList is child of CopyOnWriteArrayList. CopyOnWriteArrayList is thread safe implementation of List. The code used only public methods CopyOnWriteArrayList.size() and CopyOnWriteArrayList.get(i) to read access. 
- Possible concurrency problem case with IndexOutOfBoundsException is handled

In this implementation:
- No any garbage allocations of arrays (like toArray) or iterators (like for(v:collection_of_v))
- Case with size()==0 and size()==1 do not require special handling

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/spi/TurboFilterList.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/spi/TurboFilterList.java
@@ -39,25 +39,14 @@ final public class TurboFilterList extends CopyOnWriteArrayList<TurboFilter> {
     public FilterReply getTurboFilterChainDecision(final Marker marker, final Logger logger, final Level level,
             final String format, final Object[] params, final Throwable t) {
 
-        final int size = size();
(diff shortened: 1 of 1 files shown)


ceki:
> TurboFilterList is child of CopyOnWriteArrayList. CopyOnWriteArrayList is thread safe implementation of List. 

Yes, but TurboFilterList accesses inner (restricted) parts of CopyOnWriteArrayList. So the above assumptions no longer holds.

sgoryachkin:
> Yes, but TurboFilterList accesses inner (restricted) parts of CopyOnWriteArrayList. So the above assumptions no longer holds.

I use only public size() and public get(i) in metod getTurboFilterChainDecision(..). No any inner (restricted) parts used to iterate. TurboFilterList class has only one getTurboFilterChainDecision(..) and no any overrides of CopyOnWriteArrayList methods.

PR #610, opened 2022-11-17
fixed: keep async appender worker running when unknown exception igno…

hbiboluo:
vm options: -Xmx15m

AsyncAppenderBase#Worker will exit by OOM，main thread will be blocked forever.

```java
import org.slf4j.Logger;
import org.slf4j.MDC;

public class LoggingMain {

    private static Logger logger = LoggerFactory.getLogger(LoggingMain.class);

    protected static String getBigString(int length) {
        StringBuilder sb = new StringBuilder();
        while (sb.length() < length) {
            sb.append("#");
        }
        return sb.toString();
    }

        String bigStr = getBigString(2 * 1024 * 1024);
        while (true) {
            MDC.clear();
            logger.info("OK logger...");
            MDC.put("requestId", bigStr);
            MDC.put("userId", bigStr);
            logger.info("normal logger...");
        }
    }

}
```

```xml
<?xml version="1.0" encoding="UTF-8"?>
<configuration scan="false" scanPeriod="60 seconds" debug="true">
    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
        <layout class="ch.qos.logback.classic.PatternLayout">
            <pattern>[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%thread] [%X{requestId}] [%X{userId}] %level %logger{35} - %.-600000msg%n</pattern>
        </layout>
    </appender>
    <appender name="ASYNC_ROLLING_FILE" class="ch.qos.logback.classic.AsyncAppender">
        <discardingThreshold>0</discardingThreshold>
        <queueSize>32</queueSize>
        <appender-ref ref="STDOUT"/>
    </appender>

    <root level="DEBUG">
        <appender-ref ref="ASYNC_ROLLING_FILE" />
    </root>
</configuration>

```

```xml

```


ceki:
@hbiboluo Thank you for this PR. 
Are you getting an OutOfMemoryError in reality or is the unit test for illustrative purposes?

hbiboluo:
It has occurred several times in the production environment, then We found that this worker thread exited.

ceki:
Was the OutOfMemeoryError something that you were able to recover from? In order words, if the JVM was about to fail, does it make a difference for the worker thread to not exit? 

It is deemed bad practice to swallow Throwable hence my request for clarifications.

hbiboluo:
In our scenario, log output does not have a character length limit, and in some cases larger log output will occur.

In some extreme cases where concurrency is large, an asynchronous thread exits because it requests more array space for splicing characters. At this time, the worker thread is suspended because the queue cannot be consumed.

If the asynchronous thread ignores this log output, there is a chance that memory space will be freed up automatically when user thread processing returns. 

I agree with you that in most cases threading does not solve the problem. But the exit of the thread has lost us the chance of self-healing. Keeping threads alive is not harmful to the application. 

In a production environment, our current practice is to automatically restart container recovery services by monitoring through health checks.

PR #623, opened 2023-01-12
backport of fix LOGBACK-1162

ibrusentsev:
For fixing the LOGBACK-1162 bug that occurs when using Spring Boot 2, need to backport it to logback-1.2.x version, as Spring Boot 2 can't be upgraded to logback-1.3.x due to its upgrade policy. See https://github.com/spring-projects/spring-boot/pull/33648 for details.

(adapting commit 4d705d9045b4df494b358984d326efb2ee9d6bd0)


ceki:
It does not make sense back-port this change 1.2 and there is no need to do so.

In 1.2, you can  invoke a shutdown hook by adding the following to your config file.

`<shutdownHook class="ch.qos.logback.core.hook.DelayingShutdownHook"/>`

PR #635, opened 2023-03-15
Allows subclasses to do post-create operations on new Logger instances

rsandidfc:
Allows subclasses to do post-create operations on new Logger instances. This is very useful for subclassing Logger in conjunction with a custom context selector.

Signed-off-by: Richard Sand <rsand@idfconnect.com>


rsandidfc:
See JIRA https://jira.qos.ch/browse/LOGBACK-1747

PR #636, opened 2023-03-15
Adds additional methods to specify a custom context selector

rsandidfc:
* via system environment - especially usefor for JEE contexts
* via JNDI lookup (note this is NOT the same as using the ContextJNDISelector)

Signed-off-by: Richard Sand <rsand@idfconnect.com>

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/ClassicConstants.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/ClassicConstants.java
@@ -28,6 +28,9 @@ public class ClassicConstants {
             + "comp/env/logback/configuration-resource";
     public static final String JNDI_CONTEXT_NAME = JNDI_JAVA_NAMESPACE + "comp/env/logback/context-name";
 
+    /** JNDI name of custom contextSelector classname, if used */
+    public static final String JNDI_LOGBACK_CONTEXT_SELECTOR = "java:comp/env/logback/contextSelector";
+
     /**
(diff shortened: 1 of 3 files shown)


logback-classic/src/main/java/ch/qos/logback/classic/ClassicConstants.java
```diff
@@ -28,6 +28,9 @@ public class ClassicConstants {
             + "comp/env/logback/configuration-resource";
     public static final String JNDI_CONTEXT_NAME = JNDI_JAVA_NAMESPACE + "comp/env/logback/context-name";
 
+    /** JNDI name of custom contextSelector classname, if used */
+    public static final String JNDI_LOGBACK_CONTEXT_SELECTOR = "java:comp/env/logback/contextSelector";
```
rob93c, line 32: This constant should match existing coding conventions, just like `JNDI_CONFIGURATION_RESOURCE` and `JNDI_CONTEXT_NAME`
```suggestion
    public static final String JNDI_LOGBACK_CONTEXT_SELECTOR = JNDI_JAVA_NAMESPACE + "comp/env/logback/contextSelector";
```

rsand-temple: Yes makes total sense

rsand-temple: Should I resubmit the request?

PR #639, opened 2023-03-23
LOGBACK-1729: Add OSGi Service Loader Mediator entries to load logback Configurators

HannesWell:
At the moment logback cannot load Configurator implementations provided by other bundles in an OSGi runtime, because of missing Service Loader Mediator entries. This PR adds them, so the Service Loader Mediator implementation does not have to be configured to workaround the missing metadata.

Additionally add entry to provide the `ServletContainerInitializer` service implementation in OSGi runtimes via SPI.

In general improve the overall quality of the Service Loader Mediator entries.
See
- https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.loader.html
- https://aries.apache.org/documentation/modules/spi-fly.html#specconf

Fixes https://jira.qos.ch/browse/LOGBACK-1729.

### DIFF
--- a/logback-classic/pom.xml
+++ b/logback-classic/pom.xml
@@ -333,10 +333,17 @@
               *
             </Import-Package>
             <!-- Needed to integrate ServiceLoader mechanism with OSGi -->
-            <Require-Capability>osgi.extender; filter:="(osgi.extender=osgi.serviceloader.registrar)"</Require-Capability>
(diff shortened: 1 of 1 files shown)


HannesWell:
@ceki would that also be interesting for the logback 1.3 branch? In general the same issue exists there are well.

HannesWell:
Done in https://github.com/qos-ch/logback/pull/643.

PR #640, opened 2023-03-23
Fix OSGi execution-environment requirements and simplify configuration

HannesWell:
Remove the explicitly specified, but out dated, `Bundle-RequiredExecutionEnvironment: JavaSE-1.6` and let the maven-bundle-plugin instead add a automatically computed ee-requirement (therefore remove the `_noee` instruction). This ensures the OSGi-ExecutionEnvironment metadata are always up to date.

Also remove all other explicitly configuration that can be derived automatically and is not necessary, like more Import-Package values or `_failok`. Some of the explicitly imported packages are not even referenced in the code.


HannesWell:
@ceki would that also be interesting for the logback 1.3 branch? In general the same issue exists there are well, but the ee configuration probably has to be fine tuned a bit more due to the multi-release jar used for the modules-info.

ceki:
@HannesWell  Yes, if not too much trouble, porting the PR to the 1.3 branch would be great.

HannesWell:
> @HannesWell Yes, if not too much trouble, porting the PR to the 1.3 branch would be great.

I'm currently about to do that. Looks like the first attempt is already successful and it is less trouble than expected. :)

HannesWell:
> @HannesWell Yes, if not too much trouble, porting the PR to the 1.3 branch would be great.

It was actually straight forward. Just created https://github.com/qos-ch/logback/pull/641.

PR #642, opened 2023-03-23
LOGBACK-1730: Remove obsolete Export of package 'org.slf4j.impl' in OSGi

HannesWell:
Since SLF4J 2 the connection between SLFJ4 and its providers is established via the Service Loader mechanism, which is enabled in OSGi runtimes using a Service Loader Mediator.

The package 'org.slf4j.impl' no longer exists in this bundle and
therefore can not and need no longer be exported.

### DIFF
--- a/logback-classic/pom.xml
+++ b/logback-classic/pom.xml
@@ -318,7 +318,7 @@
         </executions>
         <configuration>
           <instructions>
-            <Export-Package>ch.qos.logback.classic*, org.slf4j.impl;version=${slf4j.version}</Export-Package>
+            <Export-Package>ch.qos.logback.classic*</Export-Package>
             <!-- LB-CLASSIC It is necessary to specify the rolling
                  file packages as classes are created via IOC (xml
                  config files). They won't be found by Bnd's analysis


HannesWell:
@ceki although this probably/hopefully does not cause problems in OSGi at the moment, because slf4j does not import the package `org.slf4j.imp`l anymore, I think it would be good to have this in 1.4 and back ported to 1.3 too.

HannesWell:
> This makes sense. Thank you for catching this. May I ask you to create a Jira issue at jira.qos.ch?

Yes I can do that. No problem.

Btw. while verifying the results of my changes by diffing against the latest 1.4.6 release I noticed that in the MANIFEST.MF the 'self-imported' packages (packages that are exported and again imported by a bundle) don't have a version range anymore, when I build logback locally. But even when I build the release commit of `v_1.4.6` I have them removed (while they are present in the jars at Maven-Central). Therefore I assume it is a (surprising) local issue on my computer. Nevertheless I would like to investigate the root of that removal first, just in case you want to perform another release soon. Nevertheless I'm puzzled why this happens.

HannesWell:
> This makes sense. Thank you for catching this. May I ask you to create a Jira issue at jira.qos.ch?

Just created https://jira.qos.ch/browse/LOGBACK-1730.

HannesWell:
> Btw. while verifying the results of my changes by diffing against the latest 1.4.6 release I noticed that in the MANIFEST.MF the 'self-imported' packages (packages that are exported and again imported by a bundle) don't have a version range anymore, when I build logback locally. But even when I build the release commit of `v_1.4.6` I have them removed (while they are present in the jars at Maven-Central). Therefore I assume it is a (surprising) local issue on my computer. Nevertheless I would like to investigate the root of that removal first, just in case you want to perform another release soon. Nevertheless I'm puzzled why this happens.

Created https://github.com/qos-ch/logback/pull/647 for that.

PR #645, opened 2023-04-03
feat: implement a json encoder

hkupty:
This is an initial implementation of a json encoder for logback.

To be noted:
- This is still work-in-progress and will be updated still, but I decided to open it earlier to enable for async discussion;
- Based on [penna's `DirectJson` implementation](https://github.com/hkupty/penna), but adapted to logback's jvm targer version and usage;

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/encoder/JsonEncoder.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/encoder/JsonEncoder.java
@@ -1,0 +1,134 @@
+package ch.qos.logback.classic.encoder;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This is a concrete JsonEncoder for {@link ILoggingEvent} that emits fields according to object's configuration.
+ * It is partially imported from <a href="https://github.com/hkupty/penna">penna</a>, but adapted to logback's structure.
+ *
+ * @author Henry John Kupty
+ */
+public class JsonEncoder extends ch.qos.logback.core.encoder.JsonEncoder<ILoggingEvent> {
+
+
+    // Excerpt below imported from
+    // ch.qos.logback.contrib.json.classic.JsonLayout
+    public static final String TIMESTAMP_ATTR_NAME = "timestamp";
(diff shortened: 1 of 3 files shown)


ceki:
@hkupty  This is a great start. Thank you.

By the way, logback commits require [Developer Certificate of Origin](https://developercertificate.org/) ](https://github.com/apps/dco) on pull requests. This is accomplished by signing commits with git's  "-s" flag. Assuming you have configured your name and email in git, no additional set up is needed.

hkupty:
Sounds good. I'll edit the commit then.

hkupty:
Do you think we can do something similar here? I'm wondering that because of the generic type `<E>` we'll need to do some sort of reflection to map the event to something we can pass along to emitters. WDYT?

I know `logback-classic` and `logback-access` have different event types, so maybe we make this an abstract class and have the respective implementations for `ILoggingEvent` and `IAccessEvent` in the appropriated packages instead.

ceki:
> Do you think we can do something similar here? I'm wondering that because of the generic type `<E>` we'll need to do some sort of reflection to map the event to something we can pass along to emitters. WDYT?

The encoder type is specified in the config file. So while, the FileAppender is event type agnostic, the encoder is not. 

So if JsonEncoderBase (for the sake of this discussion) is type agnostic, then JsonEncoder in logback-classic would know about the actual type.

> I know `logback-classic` and `logback-access` have different event types, so maybe we make this an abstract class and have the respective implementations for `ILoggingEvent` and `IAccessEvent` in the appropriated packages instead.

Yes, I would think so.

hkupty:
@ceki I've outlined what I believe to be a fair middle ground between Penna and Logback in this implementation..

Given logback is much more customizable, I believe this is the way to go if you want to use the same emitters concept in this implementation, though I'm unsure if it really suits logback general programming style.

As for the `DirectJson` implementation, I'm shaving bytes to the max in penna which makes sense given its focus and structure, but maybe we can aim to have a more "well behaved" while still having a performant implementation here where we don't need to do risky things, so it would look a bit more like [Jackson's `JsonGenerator`](https://fasterxml.github.io/jackson-core/javadoc/2.14/com/fasterxml/jackson/core/JsonGenerator.html) api, where we don't need to explicitly add field separators becasue the generator keeps track of that, for example...

How does that sound?

PR #647, opened 2023-04-07
Ensure proper version ranges of re-imported OSGi packages

HannesWell:
As mentioned in https://github.com/qos-ch/logback/pull/642#issuecomment-1481528172 when building logback locally the Imported Packages, which are also exported by a logback bundle have no version range at all. But in general they should have one.

Looking into the jars of previous releases at Maven-Central even those 're-imported' packages have a suitable version range, but I cannot really tell why they are absent in a local build. AFAICT nothing changed in this regard in the meantime and this happens is even the case if I build the logback 1.4.6 release commit locally. So I'm puzzled why this is fine in the released artifacts but isn't for my local builds.

Nevertheless this PR ensures that the imported packages, which are also exported by a logback bundle have a suitable version range.

### DIFF
--- a/logback-access/pom.xml
+++ b/logback-access/pom.xml
@@ -146,6 +146,7 @@
                 Bnd's analysis of java code.
             -->
             <Import-Package>
+              ch.qos.logback.access*;version="${range;[==,+);${maven_version;${project.version}}}",
               ch.qos.logback.core.rolling,
               ch.qos.logback.core.rolling.helper,
               org.apache.catalina.*;version="${tomcat.version}";resolution:=optional,
--- a/logback-classic/pom.xml
+++ b/logback-classic/pom.xml
@@ -324,6 +324,7 @@
                  config files). They won't be found by Bnd's analysis
                  of java code. -->
             <Import-Package>
+              ch.qos.logback.classic*;version="${range;[==,+);${maven_version;${project.version}}}",
               sun.reflect;resolution:=optional,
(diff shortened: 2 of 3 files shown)


HannesWell:
Thank you @ceki for reviewing this.
Should I also back-port this to the 1.3 branch?

ceki:
@HannesWell Yes please.

ceki:
FYI, it seems like the [version_cleanup](https://bnd.bndtools.org/macros/version_cleanup.html) macro is preferred to [maven_version](https://bnd.bndtools.org/macros/maven_version.html) macro.

HannesWell:
> FYI, it seems like the [version_cleanup](https://bnd.bndtools.org/macros/version_cleanup.html) macro is preferred to [maven_version](https://bnd.bndtools.org/macros/maven_version.html) macro.

Yes I know the documentation says `The ${version_cleanup} macro should be used in place of this macro since it can be used in more places.`, but since the `maven_version` macro works in this place and personally I found this more expressive I choose that.
But if you prefer `version_cleanup` I can create a follow up to use that, that wouldn't be a problem. :)

HannesWell:
> FYI, it seems like the [version_cleanup](https://bnd.bndtools.org/macros/version_cleanup.html) macro is preferred to [maven_version](https://bnd.bndtools.org/macros/maven_version.html) macro.

Created https://github.com/qos-ch/logback/pull/650 to use `version_cleanup`.

PR #649, opened 2023-04-09
Ensure proper version ranges of re-imported OSGi packages

HannesWell:
Back port of https://github.com/qos-ch/logback/pull/647 to the 1.3 branch.

If you prefer the `version_cleanup` macro over `maven_version`, just let me know and I'll adapt this PR accordingly (see https://github.com/qos-ch/logback/pull/647#issuecomment-1501071781).

### DIFF
--- a/logback-access/pom.xml
+++ b/logback-access/pom.xml
@@ -145,6 +145,7 @@
                 Bnd's analysis of java code.
             -->
             <Import-Package>
+              ch.qos.logback.access*;version="${range;[==,+);${maven_version;${project.version}}}",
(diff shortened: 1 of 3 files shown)


ceki:
@HannesWell `$version_cleanup` seems like a better name, but as you prefer.

HannesWell:
> @HannesWell `$version_cleanup` seems like a better name, but as you prefer.

I have no strong opinion about that. :)
Created https://github.com/qos-ch/logback/pull/650 to use version_cleanup instead.

PR #650, opened 2023-04-09
Use preferred 'version_cleanup' BND macro instead of 'maven_version'

HannesWell:
As suggested in https://github.com/qos-ch/logback/pull/647#issuecomment-1501071781.

### DIFF
--- a/logback-access/pom.xml
+++ b/logback-access/pom.xml
@@ -146,7 +146,7 @@
                 Bnd's analysis of java code.
             -->
             <Import-Package>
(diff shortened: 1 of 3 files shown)


ceki:
@HannesWell  Hi Hannes, I had to revert this commit because it seems to cause ch.qos.logback.classic.util.EnvUtilTest#versionTest to fail on the master branch, but suprisingly not on the 1.3 branch.

Something else might be going on.

ceki:
@HannesWell  I finally reverted the revert as this commit seems to be unrelated.

HannesWell:
Thanks for letting me know.

PR #652, opened 2023-04-19
Improve compatibility with 1.2 or prior

wreulicke:
I encountered unexpected behavior in the case of followings.
I think this is the compatibility issue with logback 1.2.x.
So, to improve compatibility with 1.2 or prior,
I propose default implementation returns `Instant.ofEpochMills(getTimeStamp())`.

## Reproduction

- prerequirestique
  - use logstash-logback-encoder 5.3
  - use armeria-logback 1.23.1
    -  this module is compiled with logback 1.2.x
  - use logback 1.4.x
  - logback config
```
<configuration>
    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
        <encoder class="net.logstash.logback.encoder.LogstashEncoder" />
    </appender>

    <appender name="RCEA" class="com.linecorp.armeria.common.logback.RequestContextExportingAppender">
        <appender-ref ref="STDOUT" />
    </appender>

    <root level="info">
        <appender-ref ref="RCEA" />
    </root>
</configuration>
```

The logstash-logback-encoder 7.3 calls `getInstant()` when logback 1.3 or higher, [here](https://github.com/logfellow/logstash-logback-encoder/blob/logstash-logback-encoder-7.3/src/main/java/net/logstash/logback/composite/loggingevent/LoggingEventFormattedTimestampJsonProvider.java#L30)
But, armeria-logback has LoggingEventWrapper that extends logback's LoggingEvent.
The LoggingEventWrapper lacks the implementation of `getInstant()` because it depends on logback 1.2 to keep compatibility.
Therefore, logstash-logback-encoder are going to get `null` by default implementation of `getInstant` in ILoggingEvent.
After that, logging is not working by NullPointerException.


wreulicke:
@ceki 
Thanks for your quick response.

Sorry, I don't have the jira account in `jira.qos.ch`.
Could you create a ticket?

wreulicke:
@ceki 
I created the jira ticket and updated the decription.

wreulicke:
@ceki 
Could you check this PR?

wreulicke:
@ceki 
Could you check this PR again?

PR #654, opened 2023-04-24
Make OSGi service-loader requirement for spi.Configurator optional

HannesWell:
For LOGBACK-1729 (https://jira.qos.ch/browse/LOGBACK-1729) the OSGi Service Loader Mediator Manifest entries were enhanced to be able to load logback Configurator services provided by other OSGi bundles. Unintentionally this requirement was only specified with 'cardinality:=multiple' and without 'resolution:=optional', which means that at least one bundle providing a Configurator service is required (but more are possible).

This adds the forgotten 'resolution:=optional' directive to also allow the logback-classic bundle to resolve if no Configurator service is available, which is a valid case and for what the code is prepared.

I forgot this in https://github.com/qos-ch/logback/pull/639.
@ceki can you please have a look and sorry for the extra work.


ceki:
@HannesWell  Sure can you please create a jira issue for this? The issue can be a subtask of LOGBACK-1729](https://jira.qos.ch/browse/LOGBACK-1729) if you prefer.

HannesWell:
> @HannesWell Thank you for this PR. Can you please create a jira issue for this? If you prefer, the issue can be a subtask of LOGBACK-1729](https://jira.qos.ch/browse/LOGBACK-1729).

No problem, just created [LOGBACK-1736.](https://jira.qos.ch/browse/LOGBACK-1736). I considered it as part of [LOGBACK-1729](https://jira.qos.ch/browse/LOGBACK-1729).

ceki (approved):

PR #679, opened 2023-06-21
StatusPrinter timestamp format changed to include date

jezovuk:
No JIRA issue, hopefully small enough to be accepted without one.
Added date information to timestamp format pattern in StatusPrinter for easier navigation in multi-day status log files.


jezovuk:
JIRA issue: https://jira.qos.ch/browse/LOGBACK-1751

PR #682, opened 2023-06-26
Reference the correct compression mode in the log message.

Nocera60:
Jira: LOGBACK-1750


Nocera60:
Duplicate

ceki:
@Nocera60  Funny that both PR should be so near each other while the bug has been lurking for ages...

PR #711, opened 2023-10-09
Add threadId to PatternLayout

### LINKED ISSUE #710: Add threadId to PatternLayout
Given the project already supports printing the current thread name, it should also be able to retrieve the `threadId`.

This would be useful when dealing with virtual threads, I noticed that ones created via [`Executors.newVirtualThreadPerTaskExecutor()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/Executors.html#newVirtualThreadPerTaskExecutor()) don't have a name and they would be shown as an empty string:

image

rob93c:
Fixes #710, let me know if there should be any other change I need to take care of.


rob93c:
Hi @ceki, can you please confirm if the failing Github action is doing so because of an unrelated issue?

logback-classic/src/main/java/ch/qos/logback/classic/pattern/ThreadIdConverter.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Returns the id of the current thread.
+ * 
+ * @author Roberto Cella
+ */
+public class ThreadIdConverter extends ClassicConverter {
+
+    public String convert(ILoggingEvent event) {
+        return String.valueOf(Thread.currentThread().getId());
```
griffinjm, line 26: Should this be present in the `ILoggingEvent` and retrieved from there? Otherwise this can return the thread id of an internal logback thread, not the id of the actual caller thread, e.g. in an `AsyncAppender`. See `prepareForDeferredProcessing()` in the default impl class `LoggingEvent`.

PR #769, opened 2024-01-19
Provide servlet request listener which fill the MDC

csware:
This is seen as superior to a filter as the MDC is not cleared before an error handler servlet is executed.

cf. PR #768


logback-classic/src/main/java/ch/qos/logback/classic/helpers/MDCInsertingServletRequestListener.java
```diff
@@ -0,0 +1,74 @@
...
+ * <p> The value is removed from the MDC once the request has
+ * been fully processed (including an error handler servlet).
+ *
+ * @author Sven Strickroth
+ * @author Ceki G&uuml;lc&uuml;
+ */
+public class MDCInsertingServletRequestListener implements ServletRequestListener {
+    @Override
+    public void requestInitialized(ServletRequestEvent sre) {
+        insertIntoMDC(sre.getServletRequest());
```
csware, line 39: Here, we could also cann the `insertIntoMDC` in the `MDCInsertingServletFilter` to avoid code duplication.

csware:
TODO: The documentation needs updating...

Issue #779, opened 2024-02-05
IllegalAccessException when creating thread factory


jpmsilva:
We always run our application with JDK21 and `-Dsun.reflect.debugModuleAccessChecks=true` to check for possible issues resulting from module boundary issues.
After recently upgrading logback, we started seeing this exception being logged:
```
java.lang.IllegalAccessException: class ch.qos.logback.core.util.ExecutorServiceUtil$1 cannot access a member of class java.lang.ThreadBuilders$VirtualThreadBuilder (in module java.base) with modifiers "public"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:394)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:714)
	at java.base/java.lang.reflect.Method.invoke(Method.java:571)
	at ch.qos.logback.core.util.ExecutorServiceUtil$1.makeThreadFactory(ExecutorServiceUtil.java:58)
	at ch.qos.logback.core.util.ExecutorServiceUtil$1.<init>(ExecutorServiceUtil.java:45)
	at ch.qos.logback.core.util.ExecutorServiceUtil.<clinit>(ExecutorServiceUtil.java:40)
	at ch.qos.logback.core.ContextBase.stopExecutorServices(ContextBase.java:247)
	at ch.qos.logback.core.ContextBase.stop(ContextBase.java:172)
	at ch.qos.logback.classic.LoggerContext.stop(LoggerContext.java:346)
[...snip...]
```

Now, while the class `ExecutorServiceUtil` catches that exception and handles it, it also points to an underlying issue: the `getMethod(FACTORY_METHOD_NAME)` is performed on the wrong (concrete) class `java.lang.ThreadBuilders.VirtualThreadBuilder`, instead of the public interface `java.lang.Thread.Builder`.

As such, I can't see any way the current code would work as is.

I hope to provide a PR to fix this shortly.

mkurz:
I think this can be closed now? It's part of the 1.5 release

ceki:
@mkurz Thank you for following up.

jpmsilva:
Apologies for the spam in advance.
I just wanted to confirm that the fix does indeed work, and we no longer observe the illegal access exception when starting the application with version 1.5.
[...]

PR #780, opened 2024-02-05
Fixes reflected call to construct the virtual thread factory

jpmsilva:
In reference to issue #779.


ceki:
@jpmsilva Thank you for this PR. However, `Thread.Builder` is new to java 21 so this patch is not compatible with earlier JDKs.

jpmsilva:
Ah, indeed, my apologies, I will try to provide a proper fix in a new PR soon.

jpmsilva:
Ah, I just now saw your proper fix using a multirelease jar. Thank you!

PR #790, opened 2024-03-14
Don't call file.length() for every append if we can avoid it

axiak:
This change introduces a way for the ResiliantFileOutputStream to track its own position in the file. This means that instead of calling File.length() (which ends up being a syscall) for each event, we only call it when we're recovering a stream.


ceki:
@axiak Hi Michael, 

Thank you for this contribution. When you say that  `File.length()` is called for each event, on which execution path do you think this occurs?

logback-core/src/main/java/ch/qos/logback/core/recovery/CountingOutputStream.java
```diff
@@ -0,0 +1,48 @@
+package ch.qos.logback.core.recovery;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+class CountingOutputStream extends OutputStream {
+
+  private final OutputStream delegate;
+
+  private long count;
```
griffinjm, line 10: This should probably be volatile. While there is external synchronization using various locks when calling `write()`, the new state of the variable may not be reflected immediately in other threads after this thread updates it. The increment operations `+=` and `++` are actually each 2 operations, a read and then a write, this means the `count` may be incorrectly updated by another thread afterwards, as the value it reads could be stale.

logback-core/src/test/java/ch/qos/logback/core/recovery/ResilientOutputStreamTest.java
```diff
@@ -51,15 +52,28 @@ public void verifyRecuperationAfterFailure() throws Exception {
 
         spy.write("a".getBytes());
         spy.flush();
+        assertEquals(1, spy.getCount());
 
         spy.getChannel().close();
         spy.write("b".getBytes());
         spy.flush();
+        // we have 2 in our countingoutput stream
+        // but the 'b' write failed due to the channel closing
```
griffinjm, line 61: This means the count can become inaccurate if any IOException occurs, I think it would be best to mitigate that by either only updating the count after the operation succeeds, or alternatively, if we still want to eagerly increment the count we should catch any exceptions, decrement the count, then rethrow the exception again. This would ensure the count remains accurate.

griffinjm (commented):
Accidentally clicked "start review", see my previous comments regarding the count atomicity and error handling.

griffinjm:
> @axiak Hi Michael,
> 
> Thank you for this contribution. When you say that `File.length()` is called for each event, on which execution path do you think this occurs?

From my read through the file length check occurs only when the rollover check interval is passed. For all standard `TriggeringPolicy` implementations it is not on every append.

This would be an optimization when checking for rollover if someone has a low rollover check interval. Depending on the triggering policy this can vary, for `SizeBased` it is by default 60 seconds in `DefaultInvocationGate`, but configurable, `TimeBased` are derived from the filename pattern provided. If someone has a filename date pattern which causes the policy to use a very low periodicity, e.g. MILLISECONDS or SECONDS, or someone configures the `InvocationGate` for `SizeBased` with a very low value. 

In these limited cases this could be an improvement, in exchange for extra state and complexity.

PR #804, opened 2024-04-13
fixes #803 PatternLayoutBase custom converter classloading in OSGi

### LINKED ISSUE #803: PatternLayoutBase custom converter classloading trouble in OSGi
The values in the PatternLayoutBase defaultConverterMap and instanceConverterMap are class names.  The ch.qos.logback.core.pattern.parser.Compiler class attempts to load the class for those class names with the assumption that those classes are visible to the classloader that loaded the logback-core classes.

Unfortunately, in an OSGi environment, my custom converter classes from a custom OSGi bundle are not visible to the classloader of the logback-core bundle.  This scenario results in a "Failed to instantiate converter class" error and it doesn't work as expected.

My proposal is to allow the the values of the defaultConverterMap and instanceConverterMap to alternately be supplied as a Supplier function instead of a classname.  The Supplier would not have the same class visibility problem.

Using a Supplier seems to be a technique that I see in other places in the logback codebase, so it seems like it should make sense to allow that for this scenario as well.

I will supply a PR with the proposed changes for your consideration

enapps-enorman:
This is the proposed changes to resolve issue #803

This is an alternate way to declare the PatternLayoutBase defaultConverterMap and instanceConverterMap values in a way that works with the OSGi classloading restrictions.

These changes should preserve backward compatibility with code that defines the converter map using the original technique

I also updated the tests in an attempt to confirm it works as expected.

Please review and merge if you agree with the proposed changes.

### DIFF
--- a/logback-core/pom.xml
+++ b/logback-core/pom.xml
@@ -60,6 +60,12 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>${junit-jupiter-params.version}</version>
+            <scope>test</scope>
+        </dependency>
 
         <dependency>
             <groupId>org.mockito</groupId>
--- a/logback-core/src/main/java/ch/qos/logback/core/pattern/PatternLayoutBase.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/pattern/PatternLayoutBase.java
@@ -13,8 +13,10 @@
  */
 package ch.qos.logback.core.pattern;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Supplier;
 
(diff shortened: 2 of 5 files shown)


enapps-enorman:
Hi @ceki

We would appreciate a review of this proposal if you (or someone else) has the time.  Knowing if this proposal will be accepted or not will help us decide how to move forward.

Thanks

enapps-enorman:
Hi @ceki

Checking back in again since we haven't gotten any feedback on the PR since it was proposed 4 months ago.

enapps-enorman:
re-opened...  sorry I hit the wrong button and accidentally closed it.

enapps-enorman:
Closing this PR as this problem was solved by @ceki in a different way in https://github.com/qos-ch/logback/commit/be0b5e027487c351a0f006c2b695a7a3fcd4918d and https://github.com/qos-ch/logback/commit/49f0638512cb1fc51f634ef4f972d25d34a9565b

PR #808, opened 2024-04-22
rawThrowable for JsonEncoder

zeitlinger:
add option to encode Throwable in a raw format for easy ingestion into logging backends


cyrille-leclerc:
Thanks @zeitlinger .
Would you by any chance have an example for the produced JSON payload?
To map the Logback json exception to the OpenTelemetry `exception.*`, we would need to map the following otel attributes:
* `exception.message`: The exception message. E.g. `Division by zero`; `Can't convert 'int' object to str implicitly`
* `exception.stacktrace`: A stacktrace as a string in the natural representation for the language runtime. The representation is to be determined and documented by each language SIG.
* `exception.type`: The type of the exception (its fully-qualified class name, if applicable). The dynamic type of the exception should be preferred over the static type in languages that support it. E.g. `java.net.ConnectException`

I didn't see in your code the json fields for `exception.message` and `exception.type`, did I miss something?

https://opentelemetry.io/docs/specs/semconv/attributes-registry/exception/

cyrille-leclerc:
I had in mind that the "raw stack trace" could reuse the `throwable` JSON field and just replace the `stepArray` subfield by a `stackTrace` field.

Something like:

CURRENT
```json
  "throwable": {
    "className": "java.lang.RuntimeException",
    "message": "Fraud detected",
    "stepArray": [
      {
        "className": "com.mycompany.antifraud.FraudDetectionController",
        "methodName": "checkOrder",
        "fileName": "FraudDetectionController.java",
        "lineNumber": 121
      },
     ...
      {
        "className": "java.lang.Thread",
        "methodName": "run",
        "fileName": "Thread.java",
        "lineNumber": 1583
      }
    ]
  }
```

SUGGESTION
```
  "throwable": {
    "className": "java.lang.RuntimeException",
    "message": "Fraud detected",
    "stackTrace": "| java.lang.RuntimeException: Fraud Detection Processing Exception\\n	at com.mycompany.antifraud.FraudDetectionController.checkOrder(FraudDetectionController.java:117)\\n	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)\\n	...\\n	at java.base/java.lang.Thread.run(Thread.java:1583)"
  }
```

logback-classic/src/main/java/ch/qos/logback/classic/encoder/JsonEncoder.java
```diff
@@ -264,40 +268,66 @@ private void appendThrowableProxy(StringBuilder sb, String attributeName, IThrow
         }
 
         sb.append(VALUE_SEPARATOR);
-        appendSTEPArray(sb, itp.getStackTraceElementProxyArray(), itp.getCommonFrames());
+        if (withOriginalStackTrace) {
+            appenderMember(sb, "stackTrace", getOriginalStackTrace(itp));
```
cyrille-leclerc, line 272: shouldn't `"stackTrace"` be a constant like `THROWABLE_ATTR_NAME` is a constant?

logback-classic/src/main/java/ch/qos/logback/classic/encoder/JsonEncoder.java
```diff
@@ -119,6 +122,7 @@ public class JsonEncoder extends EncoderBase<ILoggingEvent> {
     private boolean withMessage = true;
     private boolean withArguments = true;
     private boolean withThrowable = true;
+    private boolean withOriginalStackTrace = false;
```
cyrille-leclerc, line 125: Did you consider `withPlainTextStackTrace` rather than `withOriginalStackTrace` `Original` didn't resonate much with me as the `stepArray is about the same stack trace. I see he difference as a difference of format rather than a difference of data.

logback-classic/src/main/java/ch/qos/logback/classic/encoder/JsonEncoder.java
```diff
@@ -264,40 +268,66 @@ private void appendThrowableProxy(StringBuilder sb, String attributeName, IThrow
...
+    private static void getOriginalStackTrace(IThrowableProxy throwable, StringBuilder sb, int depth) {
+        if (throwable == null) {
+            return;
+        }
+        if (depth > 0) {
+            sb.append("Caused by: ");
+        }
+        sb.append(throwable.getClassName()).append(": ").append(throwable.getMessage()).append(NEWLINE);
+        for (StackTraceElementProxy step : throwable.getStackTraceElementProxyArray()) {
+            sb.append(TAB).append(step).append(NEWLINE);
```
cyrille-leclerc, line 326:
```suggestion
            sb.append(TAB).append(step.getSTEAsString()).append(NEWLINE);
```

Ensure that we use the Stack Trace Element toString even if `StackTraceElementProxy.toString()` gets modified for debugging purpose

logback-classic/src/main/java/ch/qos/logback/classic/encoder/JsonEncoder.java
zeitlinger: let's make `withPlainStackTrace`

zeitlinger:
@cyrille-leclerc please check again

zeitlinger:
@ceki I'm wondering if I found a bug here: https://github.com/qos-ch/logback/blob/bfe33c6cf03fc5fcf06cb2ba9954f71ade7fa836/logback-classic/src/main/java/ch/qos/logback/classic/encoder/JsonEncoder.java#L199

If `withThrowable` is `false`, then the the json object ends in `,}` - which is not valid.

zeitlinger:
@ceki can you review the PR?

zeitlinger:
@ceki can you review the PR?

PR #826, opened 2024-06-24
Remove logback-access from pom.xml

peterjeschke:
The module has been removed in a past release, the presence of the submodule definition made the repo not build


ceki:
Thank you for this PR. However, logback-access module is relocated. See commit 7cc62d96c30965fe which was added a few moments ago.

PR #851, opened 2024-08-24
Fix thread safety issues within BasicStatusManager

mches:
Fixes thread safety issues within `BasicStatusManager`, with respect to `count` and `level`. As it stands, updates to `count` and `level` could be temporarily or permanently invisible to other threads, or simply lost, due to lack of synchronization and non-atomic read-modify-write operations. Adopts the well-suited `LongAdder` for `count` and `AtomicInteger` for `level`. Without synchronization, `level` needs to be set using CAS to avoid a race.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/BasicStatusManager.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/BasicStatusManager.java
@@ -1,6 +1,6 @@
 /**
  * Logback: the reliable, generic, fast and flexible logging framework.
- * Copyright (C) 1999-2015, QOS.ch. All rights reserved.
+ * Copyright (C) 1999-2024, QOS.ch. All rights reserved.
  *
  * This program and the accompanying materials are dual-licensed under
  * either the terms of the Eclipse Public License v1.0 as published by
@@ -15,6 +15,8 @@ package ch.qos.logback.core;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.LongAdder;
 
 import ch.qos.logback.core.helpers.CyclicBuffer;
 import ch.qos.logback.core.spi.LogbackLock;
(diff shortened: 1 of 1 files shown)


logback-core/src/main/java/ch/qos/logback/core/BasicStatusManager.java
```diff
@@ -57,10 +59,15 @@ public void add(Status newStatus) {
...
-        }
+        count.increment();
+        int newLevel = newStatus.getLevel();
+        int currentLevel;
+        do {
+            currentLevel = level.get();
+            if (newLevel <= currentLevel) {
+                break;
+            }
+        } while (!level.compareAndSet(currentLevel, newLevel));
```
ruanwenjun, line 70: How about
```java
synchronized (statusListLock) {
      count++;
       if (newStatus.getLevel() > level) {
           level = newStatus.getLevel();
       }
       if (statusList.size() < MAX_HEADER_COUNT) {
          statusList.add(newStatus);
      } else {
          tailBuffer.add(newStatus);
      }
  }
```

The code might be more simple.

mches: Yes, we certainly could take advantage of the existing lock that's going to be taken out.

However, for correctness, the reads would also need to use the same lock:

```java
    public int getLevel() {
        synchronized (statusListLock) {
            return level;
        }
    }

    public int getCount() {
        synchronized (statusListLock) {
            return count;
        }
    }
```
Otherwise the fields would need to be made volatile:

```java
    volatile int count = 0;
    volatile int level = Status.INFO;
```

It also occurred to me that perhaps the level should be reset when the count is reset:

```java
    public void clear() {
        synchronized (statusListLock) {
            count = 0;
            level = Level.INFO;
            statusList.clear();
            tailBuffer.clear();
        }
    }
```

logback-core/src/main/java/ch/qos/logback/core/BasicStatusManager.java
```diff
@@ -57,10 +59,15 @@ public void add(Status newStatus) {
...
 
         synchronized (statusListLock) {
```
mches, line 72: Here's an alternate approach to CAS

```suggestion
        synchronized (statusListLock) {
            count.increment();
            int newLevel = newStatus.getLevel();
            if (newLevel > level.get()) {
                level.set(newLevel);
            }

```

logback-core/src/main/java/ch/qos/logback/core/BasicStatusManager.java
```diff
@@ -90,18 +97,18 @@ private void fireStatusAddEvent(Status status) {
 
     public void clear() {
         synchronized (statusListLock) {
-            count = 0;
+            count.reset();
```
mches, line 100: Should level be reset as well?

```suggestion
            count.reset();
            level.set(Status.INFO);
```

logback-core/src/main/java/ch/qos/logback/core/BasicStatusManager.java
```diff
@@ -90,18 +97,18 @@ private void fireStatusAddEvent(Status status) {
...
             statusList.clear();
             tailBuffer.clear();
         }
     }
 
     public int getLevel() {
```
mches, line 106: Should `StatusManager#getLevel()` be uncommented? 

```java
package ch.qos.logback.core.status;
…
public interface StatusManager {
…
    /**
     * Return the highest level of all the statii.
     * 
     * @return
     */
    // int getLevel();
…
}
```

Should `@Override` be added to all implemented methods?

mches:
Closing in favor of #891.

PR #869, opened 2024-10-13
Remove Unused comment in LoggerContext

ruanwenjun:


logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
```diff
@@ -132,19 +132,14 @@ private boolean isRootLogger() {
...
-                final String childName_i = childLogger_i.getName();
-
-                if (childName.equals(childName_i)) {
-                    return childLogger_i;
-                }
+        }
+        for (final Logger child : childrenList) {
+            if (childName.equals(child.getName())) {
+                return child;
             }
```
ruanwenjun, line 139: Use for-each to optimize

logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
```diff
@@ -165,11 +160,9 @@ public synchronized void setLevel(Level newLevel) {
         }
 
         if (childrenList != null) {
-            int len = childrenList.size();
-            for (int i = 0; i < len; i++) {
-                Logger child = (Logger) childrenList.get(i);
+            for (Logger childLogger : childrenList) {
                 // tell child to handle parent levelInt change
-                child.handleParentLevelChange(effectiveLevelInt);
+                childLogger.handleParentLevelChange(effectiveLevelInt);
```
ruanwenjun, line 165: Use for-each to optimize.

logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
```diff
@@ -190,10 +183,8 @@ private synchronized void handleParentLevelChange(int newParentLevelInt) {
 
             // propagate the parent levelInt change to this logger's children
             if (childrenList != null) {
-                int len = childrenList.size();
-                for (int i = 0; i < len; i++) {
-                    Logger child = (Logger) childrenList.get(i);
-                    child.handleParentLevelChange(newParentLevelInt);
+                for (Logger childLogger : childrenList) {
+                    childLogger.handleParentLevelChange(newParentLevelInt);
```
ruanwenjun, line 187: Use for-each to optimize

logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
```diff
@@ -341,23 +332,17 @@ void recursiveReset() {
         }
     }
 
-    /**
-     * The default size of child list arrays. The JDK 1.5 default is 10. We use a
-     * smaller value to save a little space.
-     */
```
ruanwenjun, line 347: Remove the unused comment.

logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
mches: I like this optimization, but I propose taking it further and ensuring that `childrenList` is always an instance of `CopyOnWriteArrayList`

```java
    transient private final List<Logger> childrenList = new CopyOnWriteArrayList<>();
```

or, with modifiers sorted

```java
    private final transient List<Logger> childrenList = new CopyOnWriteArrayList<>();
```

Then all the null checks could be removed and there would be not even a remote possibility that `childrenList`'s iterator could throw `ConcurrentModificationException`

logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
```diff
@@ -289,11 +280,11 @@ public boolean detachAppender(Appender<ILoggingEvent> appender) {
      * Create a child of this logger by suffix, that is, the part of the name
      * extending this logger. For example, if this logger is named "x.y" and the
      * lastPart is "z", then the created child logger will be named "x.y.z".
-     * 
+     *
```
mches, line 283: I'd be inclined to take out unnecessary, unrelated changes to make it easier on the maintainers

logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
ruanwenjun: Very good suggestion, I can't agree more, updated.

logback-classic/src/main/java/ch/qos/logback/classic/Logger.java
ruanwenjun: Sorry, this is caused by my IDEA setting, now I revert this.

mches (commented):
Your PR title is "Remove Unused comment in LoggerContext", but the comment you removed is in `Logger`. In `LoggerContext` you removed redundant casting and type information and unused imports. These codebase modernization activities could be done en masse using tools, but don't seem to be a project priority. I'd be inclined to separate or discard inconsequential changes. In `Logger` you do many things other than remove a comment. I suggest focusing your PR and not neglecting the PR title and description. Maintainers appreciate an accurate title and overview.

ruanwenjun:
> Your PR title is "Remove Unused comment in LoggerContext", but the comment you removed is in `Logger`. In `LoggerContext` you removed redundant casting and type information and unused imports. These codebase modernization activities could be done en masse using tools, but don't seem to be a project priority. I'd be inclined to separate or discard inconsequential changes. In `Logger` you do many things other than remove a comment. I suggest focusing your PR and not neglecting the PR title and description. Maintainers appreciate an accurate title and overview.

Thanks suggestions, this help me not only on this PR but in the future work. So I need to separate this PR and change the title to `Remove Unused comment in Logger` and focus on only remove the comment 
```java
/**
 * The default size of child list arrays. The JDK 1.5 default is 10. We use a
 * smaller value to save a little space.
 */
```

PR #879, opened 2024-10-23
Check actual file size

### LINKED ISSUE #867: SizeAndTimeBasedRollingPolicy does not take into account the initial size of an existing log file
Dependency: logback-classic:1.5.9
Java: 17
System: windows 10

Problem: when using SizeAndTimeBasedRollingPolicy and setting the maxFileSize field, if logback writes to an already existing log file (a file that was created before running the application), it will not rotate the file if the initial size is already over maxFileSize or if during execution the file exceeds maxFileSize, but it will rotate the log file only after writing maxFileSize bytes to the file, regardless of its actual size

Originally came up in issue #728 

# Reproducer

This main class generates about 1.1MB  of logs
```java
package com.example;

import org.slf4j.Logger;

public class App {

	private static final Logger LOGGER = LoggerFactory.getLogger(App.class);

		for (int i = 0; i < 20000; i++) {
			LOGGER.info("X".repeat(55));
		}
	}
}
```

logback.xml
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE configuration>

<configuration>
  <import class="ch.qos.logback.classic.encoder.PatternLayoutEncoder"/>
  <import class="ch.qos.logback.core.rolling.RollingFileAppender"/>
  <import class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy"/>

  <appender name="ROLLING" class="RollingFileAppender">
    <rollingPolicy class="SizeAndTimeBasedRollingPolicy">
      <fileNamePattern>mylog-%d{yyyy-MM-dd}.%i.txt</fileNamePattern>
      <maxFileSize>1MB</maxFileSize>
      <maxHistory>90</maxHistory>
      <totalSizeCap>20GB</totalSizeCap>
    </rollingPolicy>
    <encoder class="PatternLayoutEncoder">
      <pattern>%msg%n</pattern>
    </encoder>
  </appender>

  <root level="DEBUG">
    <appender-ref ref="ROLLING"/>
  </root>
</configuration>
```

## What happens when running the application the first time:

If there are no existing log files following the fileNamePattern: 2 log files will be generated properly

```
mylog-2024-10-11.0.txt       1025 KB
mylog-2024-10-11.1.txt       90 KB
```

## What happens when running the application multiple times after the first:

The existing log files keep getting written for `maxFileSize` bytes and no rotation occurs

```
mylog-2024-10-11.0.txt       2049 KB
mylog-2024-10-11.1.txt       179 KB
```
```
mylog-2024-10-11.0.txt       3073 KB
mylog-2024-10-11.1.txt       268 KB
```

ngocnhan-tran1996:
Closes #867

I think it's neccessary to add extra condition `activeFile.length() >= maxFileSize.getSize()` for checking the initial size of an existing log file

### DIFF
--- a/logback-classic/src/test/input/issue/logback_github867.xml
+++ b/logback-classic/src/test/input/issue/logback_github867.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <timestamp key="bySecond" datePattern="HHmm"/>
+    <appender name="ROLLING" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>logs/log_current.log</file>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <fileNamePattern>logs/log_%d{HH}.%i.log</fileNamePattern>
+            <maxFileSize>500KB</maxFileSize>
+        </rollingPolicy>
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS}  %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="ROLLING" />
+    </root>
+
+</configuration>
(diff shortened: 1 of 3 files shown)


ceki:
@ngocnhan-tran1996 Issue fixed in  85968fa3f

PR #905, opened 2025-01-09
[LOGBACK-1436] Add counter for tracking dropped and discarded events

tommyulfsparre:
This PR adds two counter that tracks the number of events that has been discarded and/or dropped. This is useful for instrumenting appenders that extends the AsyncAppenderBase.


ceki:
Hi @tommyulfsparre,

This PR fails the DCO check. Can you please fix and resubmit?

tommyulfsparre:
Hey @ceki, done!

PR #912, opened 2025-01-21
update byte count for FileAppender in prudent mode

sarveshrulz:
As this method is overridden in RollingFileAppender but was not getting called, without this the size based rolling policies don't work as lengthCounter is never updated


vogella:
We see overlapping log entries in our log even though we use prudent mode. I assume the missing update of the byte count might be the reason for this.

Issue #968, opened 2025-09-22
provide a %date{EPOCH_SECONDS}  to get the date printed as seconds from epoch


ecerulm-st:
The documentation at https://logback.qos.ch/manual/layouts.html

Currently it seems there is no way to get the epoch seconds, it delegates to  [java.time.format.DateTimeFormatter](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/time/format/DateTimeFormatter.html) and that has no way to print the seconds from the epoch. 

I think it would be useful to provide

* `%date{EPOCH_SECONDS}` like `1758543352` (integer, non fractional , no float)  (`Monday, September 22, 2025 12:15:52.005 PM` losing the milliseconds)
* `%date{EPOCH_MILLIS}` like `1758543352005` (`Monday, September 22, 2025 12:15:52.005 PM`)

This those two it will be possible to use `%date{EPOCH_SECONDS}.%date{n}` to get `1758543352.005999999` etc. 

For example I'm trying to use `logstash-logback-encoder` to get two fields in JSON `timestampSeconds` and `timestampNanos` to produce structured logs that are compliant with that the google logging agent expects ( [Logging agent: special JSON fields](https://cloud.google.com/logging/docs/structured-logging#structured_logging_special_fields) and [Time-related fields](https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing)), I can get the `timestampNanos` from `%date{n}` but I can't get the value for `timestampSeconds` from logback.  

```
       <encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
            <providers>
                <timestamp />
                <pattern>
                    <pattern>
                        {
                        "timestampSeconds": "%date{EPOCH_SECONDS}",
                        "timestampNanos":"%date{n}"
                        }
                    </pattern>
                </pattern>
                <logLevel>
                    <fieldName>severity</fieldName>
                </logLevel>

                <loggerName />
                <message />
                <mdc />
                <stackTrace />
            </providers>
        </encoder>
```

PR #1000, opened 2026-01-15
Add EpochConverter for milliseconds/seconds since epoch (related to issue #968)

duncanj:
Provides EpochConverter for including elapsed milliseconds or seconds since standard UNIX epoch (time elapsed since January 1, 1970, midnight UTC/GMT) in the log message:

* `%epoch` like `1768472130321` (for those that want milliseconds, the Java default)
* `%epoch{seconds}` like `1768472130` (for those that want seconds, the UNIX default)

I have this requirement at work, and it's been mentioned in issue #968 and in PR#969 and PR#970 which were closed as the implementation was non-optimal.

Please let me know any improvements or mistakes.  Thanks!

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/pattern/EpochConverter.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/pattern/EpochConverter.java
@@ -1,3 +1,16 @@
+/**
+ * Logback: the reliable, generic, fast and flexible logging framework.
+ * Copyright (C) 1999-2015, QOS.ch. All rights reserved.
+ *
+ * This program and the accompanying materials are dual-licensed under
+ * either the terms of the Eclipse Public License v1.0 as published by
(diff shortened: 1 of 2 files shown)


logback-classic/src/main/java/ch/qos/logback/classic/pattern/EpochConverter.java
```diff
@@ -0,0 +1,15 @@
+package ch.qos.logback.classic.pattern;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+
+public class EpochConverter extends ClassicConverter {
+    @Override
+    public String convert(ILoggingEvent event) {
+        String millisOrSeconds = getFirstOption();
```
ceki: This test should be made at the level of the start() method.

I'll make the necessary modifications.

Issue #1003, opened 2026-01-22
`NoClassDefFoundError: javax/naming/NamingException` in a modular Application with `logback.xml`


Maran23:
I have a small modular Application that uses `SLF4J` with `Logback`. 
The Application is built by using `JLink` + `JPackage`.

When running the Application, the following Error is printed and the Application will not start:
```java
Caused by: java.lang.NoClassDefFoundError: javax/naming/NamingException
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.joran.ModelClassToModelHandlerLinker.link(ModelClassToModelHandlerLinker.java:52)
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.joran.JoranConfigurator.addModelHandlerAssociations(JoranConfigurator.java:95)
	at ch.qos.logback.core@1.5.25/ch.qos.logback.core.joran.GenericXMLConfigurator.processModel(GenericXMLConfigurator.java:215)
	at ch.qos.logback.core@1.5.25/ch.qos.logback.core.joran.GenericXMLConfigurator.doConfigure(GenericXMLConfigurator.java:178)
	at ch.qos.logback.core@1.5.25/ch.qos.logback.core.joran.GenericXMLConfigurator.doConfigure(GenericXMLConfigurator.java:123)
	at ch.qos.logback.core@1.5.25/ch.qos.logback.core.joran.GenericXMLConfigurator.doConfigure(GenericXMLConfigurator.java:66)
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.util.DefaultJoranConfigurator.configureByResource(DefaultJoranConfigurator.java:68)
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.util.DefaultJoranConfigurator.configure(DefaultJoranConfigurator.java:35)
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.util.ContextInitializer.invokeConfigure(ContextInitializer.java:134)
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.util.ContextInitializer.autoConfig(ContextInitializer.java:101)
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.util.ContextInitializer.autoConfig(ContextInitializer.java:67)
	at ch.qos.logback.classic@1.5.25/ch.qos.logback.classic.spi.LogbackServiceProvider.initializeLoggerContext(LogbackServiceProvider.java:51)
[...]

PR #1004, opened 2026-01-22
Do not use javax.naming namespace in the catch block, so that Logback can be used without requiring the javax.naming module

Maran23:
Fixes: https://github.com/qos-ch/logback/issues/1003 and https://jira.qos.ch/browse/LOGBACK-1515.

The `InsertFromJNDIModelHandler` is usually created when using Logback, but only used when JNDI is used in the logging configuration. Catching the `javax.naming.NamingException` will automatically resolve the class when the `InsertFromJNDIModelHandler` class is used, throwing an exception when it is not present.

By catching the generic Exception, unless JNDI or SMTP are used, the `javax.naming` module is not required now.

The code is now similar as the `SMPTAppenderBase`, where we catch any generic `Exception` (usually the `NamingException`):

https://github.com/qos-ch/logback/blob/420d67c965b0ef76a084ac37cc45c95d16ffece3/logback-core/src/main/java/ch/qos/logback/core/net/SMTPAppenderBase.java#L141-L151

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/model/processor/InsertFromJNDIModelHandler.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/model/processor/InsertFromJNDIModelHandler.java
@@ -1,7 +1,5 @@
 package ch.qos.logback.core.model.processor;
 
-import javax.naming.NamingException;
-
 import ch.qos.logback.core.Context;
 import ch.qos.logback.core.joran.action.ActionUtil;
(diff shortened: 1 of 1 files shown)


Maran23:
> The check fails if any commit lacks a valid Signed-off-by trailer matching the commit author (usually added via **"git commit -s"**). It's a lightweight per-commit alternative to a full Contributor License Agreement (CLA), commonly enforced by bots like the DCO app.

Thanks for the info. I amended the commit with `git commit -s`. 
I also saw your response to my CLA mail! :)

PR #1026, opened 2026-03-26
fix(LOGBACK-1469): Stop orphan AsyncAppender on context reset

### LINKED ISSUE #1025: LOGBACK-1469 - Stop orphan AsyncAppender instances on context reset
## Summary
Async appenders that are **configured and started** but **never attached to any logger** are not stopped when the `LoggerContext` is reset/stopped, because shutdown walks the **logger** tree (`Logger.recursiveReset()` → `detachAndStopAllAppenders()`). Orphan appenders are not on that tree, so their **worker threads** can keep running after context teardown (e.g. Tomcat warnings about threads that failed to stop / suspected classloader leaks).
Upstream reference: [LOGBACK-1469](https://jira.qos.ch/browse/LOGBACK-1469)

## Steps to reproduce
1. Define multiple `<appender>` elements of type `AsyncAppender` with a nested `<appender-ref>`.
2. Reference **only some** of those async appenders from `<logger>` / `<root>` (leave at least one async appender **unreferenced**).
3. Start logging context, then shut down / reset the context (e.g. webapp stop).
**Observed:** `stop()` runs for appenders that were on loggers; the **unreferenced** async appender’s worker thread may still be alive
**Expected:** On context reset/stop, **every** started async appender tied to that context should be stopped and its worker thread should terminate (within the appender’s flush/stop semantics).

## Proposed direction
Register successfully started `AsyncAppenderBase` instances with the `Context` (`Context.register(LifeCycle)`) so `ContextBase.reset()` / `LifeCycleManager` stops them even when they are not reachable from any `Logger`.

samyuh:
Register AsyncAppenderBase with the context LifeCycleManager after start so ContextBase.reset() stops appenders not reachable via Logger.recursiveReset().

Tests in AsyncAppenderTest: orphan vs logger-attached scenarios.

Closes https://github.com/qos-ch/logback/issues/1025

### DIFF
--- a/logback-classic/src/test/java/ch/qos/logback/classic/AsyncAppenderTest.java
+++ b/logback-classic/src/test/java/ch/qos/logback/classic/AsyncAppenderTest.java
@@ -22,7 +22,10 @@ import ch.qos.logback.core.status.OnConsoleStatusListener;
 import ch.qos.logback.core.testUtil.RandomUtil;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
 import org.slf4j.MDC;
+
+import java.util.concurrent.TimeUnit;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


samyuh:
No longer applicable. Check https://github.com/qos-ch/logback/issues/1025 for the discussion

Issue #1031, opened 2026-04-12
Do not give deprecation warning when `converterClass`  and `class` are both specified in `conversionRule`


cholmcc:
The code _should not_ give a deprecation warning if _both_ `converterClass` and `class` tags are given in a `conversionRule` tag.  Instead, it should silently ignore the superfluous `converterClass` and use the value of `class`.   

Generally, unknown  or unsupported tags in the configuration XML should be silently ignored - certainly not raise deprecation warnings.  That is counter to the general use of XML. 

The code should, to facilitate backward compatibility, allow for _both_ tags to be present.   If a user uses older `logback`, then the same XML will cause `converterClass` to be used, while newer version will use the `class`. 

Yours,
_Christian_

ceki:
@cholmcc  Thank you for taking the time to report this bug. 

To better understand and investigate the issue, could you please provide a test case or a sample logback.xml configuration file? This would be very helpful.

Thank you again for your help.

cholmcc:
I can't - off-hand - give you are complete test-case, but I can outline it.  

Suppose you have a converter class to convert `%pid` into the process ID. 
```
public class ProcessIDConverter extends ClassicConverter {
  private static final String PID = Long.toString(ProcessHandle.current().pid());

  /** Populates %pid in log entries */
  @Override
  public String convert(final ILoggingEvent e) {
    return PID;
  }
}
```
and then you have in your `logback.xml` configuration 
```
<configuration>
  <conversionRule conversionWord="pid"
                  class="ProcessIDConverter"
                  converterClass="ProcessIDConverter" />

  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
    <file>log</file>
    <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
      <charset>UTF-8</charset>
      <pattern>%date [%pid-%thread] %-5level %logger - %msg%n</pattern>
    </encoder>
    <prudent>true</prudent>
  </appender>
  
  <root level="info">
    <appender-ref ref="FILE" />
  </root>
</configuration>
```
then `logback` will give a message _a la_ 
```
19:38:18,185 |-WARN in ch.qos.logback.core.joran.action.ConversionRuleAction - Both [converterClass] attribute and [class] attribute specified. 
``` 
which is really not a useful or meaningful message.   `logback` should simply take the value of the `class` attribute and be done with it.  

[...]

PR #1035, opened 2026-05-06
fix: do not warn when both converterClass and class attributes specified - Issue #1031

lukman48:
Fixes Issue #1031: Do not give deprecation warning when both `converterClass` and `class` are specified in `conversionRule`

## Problem
When configuring a `<conversionRule>` element in logback configuration with both legacy `converterClass` attribute and new `class` attribute (for backward compatibility), the logger issues a deprecation warning. This warning is spurious since the intention is specifically to support both attribute names for compatibility across versions.

### Current Behavior
```
WARN [ch.qos.logback.core.joran.action.ConversionRuleAction] Both [converterClass] attribute and [class] attribute specified.
WARN [ch.qos.logback.core.joran.action.ConversionRuleAction] [class] attribute will override.
```

### Expected Behavior
- No deprecation warning should be logged when both attributes are present
- Configuration with both attributes should be silently accepted
- The 'class' attribute should take precedence (as before)

## Solution
Modified `ConversionRuleAction.validPreconditions()` to:

1. **Skip deprecation warning when 'class' is present** - Only warn about 'converterClass' being deprecated if it's the ONLY class-specifying attribute present
2. **Use info-level for duplicate attributes** - Changed from WARNING to INFO level when both attributes are specified
3. **Enable backward compatibility** - Allows configuration files to include both attributes for compatibility across version upgrades

### Design Pattern
- Older logback version: uses 'converterClass' attribute (ignores 'class')
- Newer logback version: uses 'class' attribute (ignores 'converterClass')  
- Configuration with both: works seamlessly on both versions without warnings

## Implementation Details

```java
// Before: Always warned about both attributes
if(multipleClassAttributes) {
    pv.addWarn("Both [converterClass] and [class] specified.");
    pv.addWarn("[class] will override.");
}

// After: Only warn about deprecation if class alone is missing
if(!invalidConverterClassAttribute && !multipleClassAttributes) {
    pv.addWarn("[converterClass] is deprecated, use [class]");
}

if(multipleClassAttributes) {
    pv.addInfo("Both attributes specified.");  // Info, not warning
    pv.addInfo("[class] will be used.");
}
```

## Testing Checklist

- [x] Syntax validation passed (javac 17.0.17)
- [x] Only 'converterClass' present → deprecation warning shown (backward compat)
- [x] Only 'class' present → no warnings
- [x] Both attributes present → no deprecation warning, info message shown
- [x] Missing both attributes → validation error (as before)
- [x] Conversion rule works correctly with both attributes
- [x] New test case verifies both-attribute scenario
- [x] Existing tests still pass
- [x] XML configuration is properly parsed in all cases

## Files Changed

| File | Changes | Purpose |
|------|---------|---------|
| ConversionRuleAction.java | Modified validPreconditions() logic | Skip deprecation warning when both attributes present |
| PatternLayoutTest.java | Added testConversionRuleWithBothAttributes() | New test for both-attribute case |
| conversionRuleBothAttributes.xml | New test config file | Configuration with both attributes |

## Backward Compatibility
✅ Fully backward compatible - no breaking changes
✅ Deprecation warnings still shown for legacy-only configs
✅ Existing configurations continue to work unchanged

### DIFF
--- a/logback-classic/src/test/input/joran/conversionRule/conversionRuleBothAttributes.xml
+++ b/logback-classic/src/test/input/joran/conversionRule/conversionRuleBothAttributes.xml
@@ -1,0 +1,18 @@
+<configuration>
+
+  <!-- Test that when both converterClass and class attributes are specified,
+       class attribute takes precedence and no deprecation warning is issued -->
+  <conversionRule conversionWord="sample"
+    converterClass="ch.qos.logback.classic.pattern.SampleConverter"
+    class="ch.qos.logback.classic.pattern.SampleConverter" />
+
+  <appender name="LIST" class="ch.qos.logback.core.testUtil.StringListAppender">
+    <layout class="ch.qos.logback.classic.PatternLayout">
+      <Pattern>%sample - %msg</Pattern>
+    </layout>
+  </appender>
+
+  <root level="debug">
+    <appender-ref ref="LIST" />
+  </root>
+</configuration>
--- a/logback-classic/src/test/java/ch/qos/logback/classic/PatternLayoutTest.java
+++ b/logback-classic/src/test/java/ch/qos/logback/classic/PatternLayoutTest.java
@@ -285,6 +285,21 @@ public class PatternLayoutTest extends AbstractPatternLayoutBaseTest<ILoggingEve
     }
 
     @Test
+    public void testConversionRuleWithBothAttributes() throws JoranException {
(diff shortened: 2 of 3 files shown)


ceki:
@lukman48 Thank you for this PR. It looks good to me. May I ask which model and coding agent you used for this PR?

lukman48:
I use github copilot with model Claude haiku 4.5

PR #1036, opened 2026-05-07
Fix #1001: Suppress spurious warning when including classpath config files without scan

### LINKED ISSUE #1001: including file from classpath produces unavoidable warning?
Hi,
having a `logback.xml` on classpath, using `<include resource="otherfile.xml"/>` element, logback always warns about:

`17:11:20,313 |-WARN in ch.qos.logback.core.joran.util.ConfigurationWatchListUtil@1dd6d4b7 - Null ConfigurationWatchList. Cannot add file:/C:/dev/CTSS-PoC.git/vass-supervisor/gateway/target/classes/_logback-monitor.inc.xml
`

There are no other warnings, and there is no way to avoid/workaround this warning. 
Tried with <configuration scan="true">, but that does not work (and makes no sense, right) with resources from classpath.
The included file is existing and loaded and the logging context is working as expected. There is no reason for logback showing anything in `StatusPrinter2.printInCaseOfErrorsOrWarnings()`.

The expected behaviour is: show no WARN at all.

Can you suppress this warning above, or make it conditional showing only if warching is enabled at all?

lukman48:
## Summary
When a resource is included via `<include resource="file.xml"/>` without `scan="true"`, logback would log an unnecessary INFO message about ConfigurationWatchList not being initialized. This produces spurious warnings in logs even though everything works correctly.

## Root Cause
The `addToWatchList()` method was logging a message whenever ConfigurationWatchList was null, regardless of whether file monitoring was actually enabled.

## Solution
Only log the message when `createCWL=true` (which indicates that an attempt was made to create a watch list). When `createCWL=false` (the default for `<include>` elements), silently return without logging, as file monitoring is not enabled and the classpath resource cannot be monitored anyway.

## Changes
- Modified `ConfigurationWatchListUtil.addToWatchList()` to conditionally log based on createCWL parameter
- Added comprehensive test class `ConfigurationWatchListUtilTest` with 4 test cases covering:
  - No spurious messages when createCWL=false (Issue #1001)
  - Proper watch list creation when createCWL=true with watchable protocol
  - Existing watch list behavior remains unchanged
  - Default behavior validation

## Benefits
✅ Cleaner logs: no spurious messages when `scan="false"` (the default)  
✅ Expected behavior: resources from classpath without scan directive don't log warnings  
✅ Maintains backward compatibility: existing watch list behavior unchanged  
✅ All existing tests pass  

## Testing
- New test class: `ConfigurationWatchListUtilTest.java` (4 test cases)
- Existing test suite passes: `IncludeModelHandlerTest` validates includes work correctly
- Full `logback-core` test suite passes

Fixes #1001

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/joran/util/ConfigurationWatchListUtil.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/joran/util/ConfigurationWatchListUtil.java
@@ -95,7 +95,11 @@ public class ConfigurationWatchListUtil {
             if(createCWL && ConfigurationWatchList.isWatchableProtocol(url)) {
                 cwl = registerNewConfigurationWatchListWithContext(context);
             } else {
-                addInfo(context, "ConfigurationWatchList not initialized due to absence of scan directive. Will not add " + url);
+                // Only log if explicitly trying to create a watch list and it's a non-watchable protocol
+                // Otherwise, silently return when watching is not enabled (scan=false)
+                if(createCWL) {
+                    addInfo(context, "ConfigurationWatchList not initialized due to absence of scan directive. Will not add " + url);
+                }
                 return;
             }
         }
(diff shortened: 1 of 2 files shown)


lukman48:
Closing in favor of updated PR with proper DCO sign-off.

PR #1045, opened 2026-07-02
Fix typo in Throwable class name

t0rchwo0d:
Hi, I noticed a typo in the `JAVA_CLASSES` whitelist array.

In the current implementation, `Throwable` is written with a comma instead of a period, which prevents `java.lang.Throwable` from being correctly recognized or filtered.

* **Before**: `"java,lang.Throwable"`
* **After**: `"java.lang.Throwable"`

Thank you!


ceki:
Part of release 1.5.38.

PR #1051, opened 2026-07-14
Support HALF_DAY periodicity for AM/PM date patterns

### LINKED ISSUE #976: TimeBasedRollingPolicy does not honor HALF_DAY periodicity for pattern %d{yyyy-MM-dd-a} — rolls over only at midnight instead of 00:00 and 12:00
When using a TimeBasedRollingPolicy with a date pattern containing the AM/PM marker (a), such as:

`<fileNamePattern>/var/log/app.log.%d{yyyy-MM-dd-a}</fileNamePattern>
`

Logback initializes with the message:

```
The date pattern is 'yyyy-MM-dd-a'.
Roll-over at midnight.
```

and performs daily rotation only (00:00), not twice per day (00:00 and 12:00) as expected.

**Expected behavior**

For a pattern including the AM/PM token **(a)**, Logback should detect **HALF_DAY** periodicity and rotate logs at midnight and noon.

**Actual behavior**

Startup message says “Roll-over at midnight.”

Only one rollover per day occurs.

The enum constant `HALF_DAY `exists in `RollingCalendar (printPeriodicity()` even prints `"Roll-over at midday and midnight.")`, but this value is never selected by the periodicity detector.

**Technical cause**

From source inspection:

`RollingCalendar#printPeriodicity()` supports `HALF_DAY`.

However, `RollingCalendar#innerGetEndOfNextNthPeriod()` and the periodicity detector never implement logic for `HALF_DAY`, so detection always falls back to `TOP_OF_DAY`.

As a result, `DefaultTimeBasedFileNamingAndTriggeringPolicy` always logs `“Roll-over at midnight.”`

This behavior is reproducible in Logback 1.4.x (tested with 1.4.14) and appears to match the old [LOGBACK-1552 report](https://jira.qos.ch/browse/LOGBACK-1552) (“yyyy-MM-dd-a not working as expected”).

**Steps to Reproduce**

Configure:

```
<appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
  <file>app.log</file>
  <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
    <fileNamePattern>app.log.%d{yyyy-MM-dd-a}</fileNamePattern>
  </rollingPolicy>
</appender>
```

Run an app that logs periodically.

Observe startup message and verify only one rotation at 00:00.

**Proposed fix**

Add explicit `HALF_DAY `handling in:

`RollingCalendar#innerGetEndOfNextNthPeriod()`

`RollingCalendar#getPeriodicityType(…)` (so `%d{…a}` correctly maps to HALF_DAY)
to align detection with `printPeriodicity()`’s existing message and enable true twice-daily rotation.

seonwooj0810:
Fixes #976

## Root cause
The `HALF_DAY` periodicity type and its `RollingCalendar.printPeriodicity()` message ("Roll-over at midday and midnight.") have existed since LBCORE-11 (2008), but `HALF_DAY` was **never** added to `PeriodicityType.VALID_ORDERED_LIST`, and `RollingCalendar` has no `HALF_DAY` branch in its switch statements. So `computePeriodicityType()` can never select it: a `TimeBasedRollingPolicy` whose date pattern carries the AM/PM token but no finer time token (e.g. `%d{yyyy-MM-dd-a}`) falls through to `TOP_OF_DAY` and rolls over only at midnight, logging "Roll-over at midnight." instead of rotating at both 00:00 and 12:00.

## Change
- `PeriodicityType`: insert `HALF_DAY` into `VALID_ORDERED_LIST` between `TOP_OF_HOUR` and `TOP_OF_DAY`. Because detection returns the shortest matching period, this only affects patterns that distinguish 00:00 from 12:00 but nothing finer (i.e. the `a` token) — patterns that already match at `TOP_OF_HOUR` or below are unchanged.
- `RollingCalendar.innerGetEndOfNextNthPeriod`: add the `HALF_DAY` case — floor to the current half-day boundary (00:00 or 12:00) and advance in 12h steps, mirroring the existing `TOP_OF_DAY` flooring.
- `RollingCalendar.periodBarriersCrossed`: add `HALF_DAY` (previously it hit the `default` branch and threw `IllegalStateException`).
- `RollingCalendar.isCollisionFree`: add `HALF_DAY` — an isolated `a` without a date component repeats every 24h, so a 1-day collision check flags it.

## Test evidence
Added regression coverage to `RollingCalendarTest` (all pass; each fails on current `master` before the fix):
- `testPeriodicity`: `yyyy-MM-dd-a` and `yyyy-MM-dd a` now resolve to `HALF_DAY` (previously `TOP_OF_DAY`).
- `testVaryingNumberOfHalfDailyPeriods`: `getEndOfNextNthPeriod` advances in exact 12h boundaries across a range of periods.
- `testBarrierCrossingComputation`: 3 half-day barriers over 36h.
- `testCollisionFreenes`: `yyyy-MM-dd-a` is collision-free; isolated `a` is not.

Verification done: `mvn -pl logback-core -am test -Dtest=RollingCalendarTest` → 10/10 pass. Ran the full `ch.qos.logback.core.rolling.*` suite: all rolling tests pass; the only errors are 3 pre-existing `InaccessibleObjectException`s in `RollingFileAppenderTest` (JPMS reflection into `ch.qos.logback.core.appender`), which are unrelated to this change and reproduce identically on a clean `master` checkout when the test JVM is JDK 25.

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/rolling/helper/PeriodicityType.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/rolling/helper/PeriodicityType.java
@@ -21,6 +21,7 @@ public enum PeriodicityType {
     // The followed list consists of valid periodicity types in increasing period lengths
     static PeriodicityType[] VALID_ORDERED_LIST = new PeriodicityType[] { TOP_OF_MILLISECOND,
             PeriodicityType.TOP_OF_SECOND, PeriodicityType.TOP_OF_MINUTE, PeriodicityType.TOP_OF_HOUR,
-            PeriodicityType.TOP_OF_DAY, PeriodicityType.TOP_OF_WEEK, PeriodicityType.TOP_OF_MONTH };
+            PeriodicityType.HALF_DAY, PeriodicityType.TOP_OF_DAY, PeriodicityType.TOP_OF_WEEK,
+            PeriodicityType.TOP_OF_MONTH };
 
 }
--- a/logback-core/src/main/java/ch/qos/logback/core/rolling/helper/RollingCalendar.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/rolling/helper/RollingCalendar.java
@@ -105,6 +105,10 @@ public class RollingCalendar extends GregorianCalendar {
             // isolated hh or KK
             return !collision(12 * MILLIS_IN_ONE_HOUR);
 
+        case HALF_DAY:
+            // isolated 'a' (AM/PM) without a date component repeats every day
+            return !collision(MILLIS_IN_ONE_DAY);
+
         case TOP_OF_DAY:
             // EE or uu
             if (collision(7 * MILLIS_IN_ONE_DAY))
@@ -198,6 +202,8 @@ public class RollingCalendar extends GregorianCalendar {
             return diff / MILLIS_IN_ONE_MINUTE;
         case TOP_OF_HOUR:
             return diff / MILLIS_IN_ONE_HOUR;
+        case HALF_DAY:
+            return diff / (12 * MILLIS_IN_ONE_HOUR);
         case TOP_OF_DAY:
             return diff / MILLIS_IN_ONE_DAY;
         case TOP_OF_WEEK:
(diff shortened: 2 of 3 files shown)


seonwooj0810:
Checking in on this feature whenever you get a chance — open to feedback if you'd rather shape the API differently.

PR #1055, opened 2026-07-18
Fix data race on LoggerContext.size (#1038)

### LINKED ISSUE #1038: Data race on LoggerContext.size
[LoggerContext.size](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/LoggerContext.java#L57) has a data race. [getLogger()](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/LoggerContext.java#L156) synchronizes on the current logger tree node,
but concurrent callers from different threads hold different monitors, so [incSize()](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/LoggerContext.java#L161) → [size++](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/LoggerContext.java#L172)
is unprotected. ThreadSanitizer reports:

```
==================
WARNING: ThreadSanitizer: data race (pid=246682)
  Write of size 4 at 0x0006c0043a10 by thread T1908 (mutexes: write M0):
    #0  ch.qos.logback.classic.LoggerContext.incSize()         LoggerContext.java:167
    #1  ch.qos.logback.classic.LoggerContext.getLogger(String) LoggerContext.java:156
    #2  ch.qos.logback.classic.LoggerContext.getLogger(String) LoggerContext.java:53
    #3  org.slf4j.LoggerFactory.getLogger(String)              LoggerFactory.java:363
    #4  org.slf4j.LoggerFactory.getLogger(Class)               LoggerFactory.java:388
    #5  ... 

  Previous write of size 4 at 0x0006c0043a10 by thread T1915 (mutexes: write M1):
    #0  ch.qos.logback.classic.LoggerContext.incSize()         LoggerContext.java:167
    #1  ch.qos.logback.classic.LoggerContext.getLogger(String) LoggerContext.java:156
    #2  ch.qos.logback.classic.LoggerContext.getLogger(String) LoggerContext.java:53
    #3  org.slf4j.LoggerFactory.getLogger(String)              LoggerFactory.java:363
    #4  org.slf4j.LoggerFactory.getLogger(Class)               LoggerFactory.java:388
    #5  ... 
==================
```

seonwooj0810:
Fixes #1038

### Root cause
`LoggerContext.getLogger(String)` creates each new logger under the monitor of its **parent** node (`synchronized (logger)`). Concurrent callers creating loggers below *different* parents therefore hold *different* locks, yet they all update the same counter via `incSize()`, which performed an unsynchronized `size++`. That read-modify-write races, so increments are lost under contention and `size()` reports fewer loggers than actually exist. This is exactly the ThreadSanitizer data race reported in the issue.

### Change
Make the counter an `AtomicInteger` and use `incrementAndGet()` / `get()`, so the count is maintained atomically regardless of which parent monitor a thread happens to hold. No locking is added to the hot path.

### Test evidence
Added `LoggerContextTest#concurrentGetLoggerKeepsSizeConsistent`: 16 threads each create 500 loggers under their own distinct parent (so their `incSize()` calls hold distinct monitors, maximising contention), then the test asserts `size()` matches the real logger count.

- Before the fix the test fails, e.g. `expected: <8017> but was: <7970>` (lost increments).
- After the fix it passes; the full `LoggerContextTest` (17 tests) is green.

Verification done: built and ran `mvn -pl logback-classic -am -Dtest=LoggerContextTest test` on the branch (all pass), and confirmed the new test fails on unpatched `master`.

### DIFF
--- a/logback-classic/src/main/java/ch/qos/logback/classic/LoggerContext.java
+++ b/logback-classic/src/main/java/ch/qos/logback/classic/LoggerContext.java
@@ -34,6 +34,7 @@ import org.slf4j.spi.MDCAdapter;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static ch.qos.logback.core.CoreConstants.EVALUATOR_MAP;
 
@@ -54,7 +55,10 @@ public class LoggerContext extends ContextBase implements ILoggerFactory, LifeCy
     public static final boolean DEFAULT_PACKAGING_DATA = false;
 
     final Logger root;
-    private int size;
+    // loggers are created under per-parent monitors in getLogger(String), so
+    // concurrent creations run under different locks; the shared logger count
+    // must therefore be updated atomically to avoid a data race (see #1038).
+    private final AtomicInteger size = new AtomicInteger();
     private int noAppenderWarning = 0;
     final private List<LoggerContextListener> loggerContextListenerList = new ArrayList<LoggerContextListener>();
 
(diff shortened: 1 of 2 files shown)


seonwooj0810:
Following up on this data-race fix — happy to adjust the synchronization approach or add more tests if needed.

PR #349, opened 2016-12-13
closing DatagramSocket to ensure underlying FileDescriptor is closed …

barney2k7:
…immediately

fixes [LOGBACK-1019](http://jira.qos.ch/browse/LOGBACK-1019)

PR #230, opened 2014-12-17
[LOGBACK-1019] Added code to close the DatagramSocket

nrkkalyan:
I am surprised to see there is no close method called on the DatagramSocket this may lead to a UDP leak.
Kindly let me know if there is anything wrong in my code.

PR #1057, opened 2026-07-21
LOGBACK-1019: SyslogOutputStream.close() never closes its DatagramSocket

aysha-afrah26:
`SyslogOutputStream.close()` nulls the DatagramSocket reference but never calls close() on it, so the FD leaks until GC finalization. one socket per appender stop (e.g. config reload with scan="true") and per ResilientSyslogOutputStream recovery cycle, eventually leading to `SocketException: maximum number of DatagramSockets reached` (see comments on #230).
Same fix as #230 / #349, which have been open for years — but this one includes a regression test (fails on master, passes with the fix).


ceki:
@aysha-afrah26 Thank you for this PR. Fixed in https://github.com/qos-ch/logback/commit/159c045d8f045ccf8b382775d81d83919c228cca.

PR #1064, opened 2026-08-11
Do not close console stream on ConsoleAppender.stop

### LINKED ISSUE #1063: ConsoleAppender with `withJansi=true` closes the JVM's stdout file descriptor on `LoggerContext.reset()`
## Summary

With `<withJansi>true</withJansi>` on a `ConsoleAppender`, calling `LoggerContext.reset()` (or `stop()`) closes the JVM's **stdout file descriptor**, not merely the appender's own stream. Everything writing to stdout afterwards is silently lost, and the condition is not recoverable within the JVM.

Originally reported at [LOGBACK-1759](https://jira.qos.ch/browse/LOGBACK-1759).

## Environment

logback-classic / logback-core 1.3.16, slf4j-api 2.0.18, jansi 2.4.3, Java 8, Windows.
(The same code path exists unchanged on current `master`.)

## Analysis

1. `LoggerContext.reset()` → `Logger.recursiveReset()` → `detachAndStopAllAppenders()` → `Appender.stop()`
2. `OutputStreamAppender.stop()` calls `closeOutputStream()`, which does `this.outputStream.close()`.
3. Without Jansi this is harmless by design: `ConsoleAppender.start()` uses `ConsoleTarget.SystemOut.getStream()`, an anonymous `OutputStream` that overrides only `write(..)` and `flush()`. Its `close()` is inherited from `java.io.OutputStream` and is a no-op. That is the existing safety net.
4. With `withJansi=true` that safety net is bypassed. `ConsoleAppender.wrapWithJansi()` takes the Jansi-2 branch and **discards `targetStream` entirely**:

   ```java
   // check for JAnsi 2
   if (optOutMethod.isPresent()) {
       return (PrintStream) outMethod.invoke(null);   // == AnsiConsole.out()
   }
   // JAnsi 1
   return (OutputStream) method.invoke(null, new PrintStream(targetStream));
   ```

   Only the Jansi-1 fallback still wraps the non-closing `ConsoleTarget` stream.

5. `AnsiConsole.out()` returns the process-wide singleton built by `ansiStream(true)`:

   ```java
   FileDescriptor descriptor = stdout ? FileDescriptor.out : FileDescriptor.err;
   final OutputStream out = new FastBufferedOutputStream(new FileOutputStream(descriptor));
   ```

   i.e. it sits directly on `FileDescriptor.out`, *not* on `System.out`.

6. `AnsiPrintStream` does not override `close()`. So `PrintStream.close()` → `AnsiOutputStream.close()` (`uninstall(); super.close();`) → `FileOutputStream(FileDescriptor.out).close()` → fd 1 is closed at OS level.

## Why reconfiguration does not repair it

`AnsiConsole.out` is a static field whose recreation is guarded by the `initialized` flag in `initStreams()`, and `close()` does not clear that flag. A subsequent `JoranConfigurator.doConfigure(..)` therefore hands the new `ConsoleAppender` back the **same, already closed** `AnsiPrintStream`. `System.setOut(..)` does not help either, because the descriptor itself is gone.

## Reproduction

`logback.xml` with a single `ConsoleAppender` using `<withJansi>true</withJansi>`, then:

```java
LoggerContext lc = (LoggerContext) LoggerFactory.getILoggerFactory();
LoggerFactory.getLogger("x").info("hello");
System.out.println("before reset");               // visible
System.out.println(FileDescriptor.out.valid());   // true
lc.reset();
System.out.println("after reset");                // lost
System.err.println(FileDescriptor.out.valid());   // false
```

## Practical impact

Test suites that reconfigure logback per test (JUnit/ScalaTest `@Before`/`@After`) must call `reset()` to avoid appender duplication on re-configuration. Doing so kills stdout for the rest of the run; with an unforked build tool it also kills the build tool's own console output, so the run aborts rather than merely losing log lines.

## Suggested fix

`ConsoleAppender` should not close a stream it does not own. The most direct fix is to override `closeOutputStream()` in `ConsoleAppender` to flush only - this mirrors `java.util.logging.ConsoleHandler.close()`, which explicitly flushes without closing `System.err`, and it makes the `withJansi` case behave like the already-safe `ConsoleTarget` case. Alternatively, wrap the stream returned by `AnsiConsole.out()` in a non-closing delegate.

Note that `master` is still affected and additionally invokes `AnsiConsole.systemInstall()` in `wrapWithJansi()`, which makes the closed singleton the process-wide `System.out` as well.

## Workarounds

- Set `withJansi` to `false`. On Windows 10 1511+ the console interprets VT sequences natively, and build tools that install Jansi themselves keep colouring working through `ConsoleTarget` → `System.out`.
- Or use a `ConsoleAppender` subclass overriding `closeOutputStream()` to flush only.

kalayciburak:
## Summary
`ConsoleAppender.stop()` inherited `OutputStreamAppender.closeOutputStream()`, which calls `close()` on the appender's stream. That is safe for the default `ConsoleTarget` streams (their `close()` is a no-op), but with `<withJansi>true</withJansi>` Jansi 2 returns a process-wide `AnsiPrintStream` sitting on `FileDescriptor.out`. Closing it permanently shuts down the JVM's stdout FD, so later `System.out` writes and even reconfiguration are lost.

## Fix
Override `closeOutputStream()` in `ConsoleAppender` to write the encoder footer and **flush** only — same contract as `java.util.logging.ConsoleHandler.close()`.

Fixes #1063 (LOGBACK-1759).

## Test plan
- [x] `mvn -pl logback-core test -Dtest=ConsoleAppenderCloseTest` (2/2) — RED before fix (stream closed), GREEN after
- [x] `mvn -pl logback-core,logback-core-blackbox -am test -Dtest=ConsoleAppenderCloseTest,OutputStreamAppenderTest,JlineJansiConsoleAppenderTest,FuseSourceJansiConsoleAppenderTest` (all GREEN)

### DIFF
--- a/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
+++ b/logback-core/src/main/java/ch/qos/logback/core/ConsoleAppender.java
@@ -13,6 +13,7 @@
  */
 package ch.qos.logback.core;
 
+import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.lang.reflect.Method;
@@ -22,6 +23,7 @@ import java.util.NoSuchElementException;
 import java.util.Optional;
 
 import ch.qos.logback.core.joran.spi.ConsoleTarget;
+import ch.qos.logback.core.status.ErrorStatus;
 import ch.qos.logback.core.status.Status;
 import ch.qos.logback.core.status.WarnStatus;
 import ch.qos.logback.core.util.Loader;
@@ -239,6 +241,28 @@ public class ConsoleAppender<E> extends OutputStreamAppender<E> {
      */
     public void setWithJansi(boolean withJansi) {
         this.withJansi = withJansi;
+    }
+
+    /**
+     * Flush encoder footer and the underlying stream, but do not close it.
+     * <p>
+     * {@link ConsoleAppender} does not own {@code System.out}/{@code System.err}, nor the
+     * process-wide Jansi {@code AnsiPrintStream} returned when {@code withJansi} is enabled.
+     * Closing that stream (Jansi 2 sits on {@link java.io.FileDescriptor#out}) shuts down the
+     * JVM's stdout file descriptor permanently — see LOGBACK-1759 / issue #1063. This mirrors
+     * {@code java.util.logging.ConsoleHandler#close()}, which flushes without closing the console.
+     */
+    @Override
+    protected void closeOutputStream() {
+        if (getOutputStream() != null) {
+            try {
+                // write encoder footer before detaching from the stream
+                encoderClose();
+                getOutputStream().flush();
+            } catch (IOException e) {
+                addStatus(new ErrorStatus("Could not flush output stream for ConsoleAppender.", this, e));
+            }
+        }
     }
 
 }
--- a/logback-core/src/test/java/ch/qos/logback/core/ConsoleAppenderCloseTest.java
+++ b/logback-core/src/test/java/ch/qos/logback/core/ConsoleAppenderCloseTest.java
@@ -1,0 +1,101 @@
+/*
+ * Logback: the reliable, generic, fast and flexible logging framework.
+ * Copyright (C) 1999-2026, QOS.ch. All rights reserved.
+ *
+ * This program and the accompanying materials are dual-licensed under
+ * either the terms of the Eclipse Public License v2.0 as published by
+ * the Eclipse Foundation
+ *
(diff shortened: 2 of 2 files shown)


ceki:
@kalayciburak Thank you for this PR.

However,  given https://github.com/qos-ch/logback/issues/1063 I had already started working on this topic. I think commits ad47c2ef7a1ef55 and subsequent commits largely solved the problem you are raising.

ceki:
I am marking this PR as duplicate and closing.

